Issue #15, opened 2026-04-18
Simulator: coroutine-aware async driver so Lua can await JS Promises


bvdaakster:
## Problem

Wasmoon returns JavaScript Promises to Lua as opaque userdata. The Lua code has no way to await them, so any simulator mock backed by an async-only browser API (`DecompressionStream`, `fetch`, WebBluetooth, File System Access, `indexedDB` cursors, etc.) cannot be plumbed through a C-equivalent binding without either blocking the UI thread or being skipped outright.

Today we dodge this with synchronous substitutes: sync XHR for range reads in `mock/storage.js`, pako for `ez.compression.inflate` in `mock/compression.js`. That's fine for the current surface area, but new bindings will keep colliding with this wall.

## Proposal

Implement the same pattern the firmware already uses for `async_read_bytes` — **a coroutine-yielding bridge**.

### Lua side

Screens already run Lua through `spawn(fn)` which wraps `fn` in a coroutine. Add a helper:

```lua
-- ezui.async.js_await(js_promise) -> value, or raises on rejection
-- Must be called from inside a spawn()-ed coroutine.
function async.js_await(promise)
    local co = coroutine.running()
    assert(co, "js_await must run inside a coroutine")
    local done, value, err = false, nil, nil
    _js_bridge.attach(promise,
        function(v) value = v; done = true; _js_bridge.resume(co) end,
        function(e) err   = e; done = true; _js_bridge.resume(co) end)
    while not done do coroutine.yield() end
    if err then error(err) end
    return value
end
```

### JS side

Extend `simulator.js` with a small bridge that the C-equivalent bindings call:

```js
const pending = new Set();
lua.global.set('_js_bridge', {
    attach: (promise, onResolve, onReject) => {
        promise.then(onResolve, onReject);
    },
    resume: (co) => {
        pending.add(co);
        // Drain queued coroutines on the next tick.
        Promise.resolve().then(drainPending);
    },
});

function drainPending() {
    for (const co of pending) {
        try { lua.global.get('coroutine').resume(co); }
        catch (e) { logError(e); }
    }
    pending.clear();
}
```

### Bindings

Mocks that wrap an async API can then do:

```js
createFileModule(lua) {
    return {
        async_read(path) {
            return fetch(path).then(r => r.text());
        },
    };
}
```

and the Lua caller writes `ez.async.js_await(ez.file.async_read(path))` — which from the screen's perspective is indistinguishable from the firmware's `async_read_bytes`.

## Acceptance

[...]

Issue #14, opened 2026-04-18
Deprecate RLE tile compression in a future TDMAP format bump


bvdaakster:
## Background

As of the v5 format bump + zlib rollout, new archives default to `COMPRESSION_ZLIB` (flag byte `2`). Measured on the Netherlands source:

- RLE:     1.4–1.9× compression ratio
- Zlib(9): 3.4–11.6× compression ratio

Amsterdam sample: 2.7 MB → 994 KB (~63% smaller). Extrapolated to the world archive: ~478 MB → ~80 MB. Decoding is handled by the ROM miniz on-device (`tinfl_decompress_mem_to_mem`, sub-millisecond per tile) via `ez.compression.inflate`, so decoded cost is lower than the old Lua RLE path.

Current readers accept both v4 (implicit RLE) and v5 (compression byte dispatched). That is the right compatibility window for now — users with SD cards full of v4 archives shouldn't need to rebuild.

## Proposal

When the format next bumps (v6 or later), drop RLE support entirely:

- Writer: remove the `COMPRESSION_RLE` code path in `process.py` / `archive.py`.
- Reader (Python `TDMAPReader`): raise on `compression != COMPRESSION_ZLIB` for v6+.
- Reader (Lua `services/map_archive`): drop the `rle_decompress` fallback for v6+. Keep it gated behind `version <= 5` so legacy archives still open.
- Remove `COMPRESSION_RLE` and `rle_compress`/`rle_decompress` from `tools/maps/process.py`.
- Document the deprecation in the README (migration path: rebuild archives).

## Why wait

- The map_archive Lua service is only ~280 lines; removing the RLE branch now saves almost nothing.
- Users who hand-copied `world.tdmap` before the zlib landing will have RLE archives on SD. Forcing them to rebuild with no warning is rude.
- The v4→v5 migration gave us the header byte we needed to dispatch, so this is a purely removal-only change when we're ready.

## Trigger to pull the plug

Any of:
- Another format change is landing anyway (bundle the RLE removal with it).
- Six months have passed since v5 shipped.
- Someone actively needs the code simplification for a new feature.

bvdaakster:
Done across two commits on `feature/offline-maps-v5-zlib`:

- **4457967** dropped the v4/v5 reader paths entirely (writer was already zlib-only since v6). The on-device Lua reader rejects non-v6 archives at open time with a "regenerate with the current writer" message; the Python writer/reader only emits and accepts v6.
[...]

PR #16, opened 2026-04-18
Offline maps + shell UI overhaul: status bar, glass icons, AA fonts, SND01 sounds

### LINKED ISSUE #2: tools/maps: port viewer.html to TDMAP v4 geographic labels
## Problem

`tools/maps/viewer.html` still parses the v3 tile-indexed label format. TDMAP was bumped to v4 (geographic lat/lon labels, sequential layout, no per-tile index) in commit `38add06`. This means archives built today cannot be verified in the desktop viewer, which is the main sanity-check step before flashing to the device.

## Proposal

- Mirror `TDMAPReader.get_labels_in_bounds()` (see `tools/maps/archive.py`) in JS.
- Parse the v4 sequential label block: fixed-size header + variable-length UTF-8 text.
- Filter by visible viewport (lat/lon bounds) and current zoom against each label's `zoom_min`/`zoom_max`.
- Keep backwards compatibility with v3 archives, or drop it if we are confident no v3 `.tdmap` files remain in the wild.

## Acceptance

- Loading a v4 archive in `viewer.html` shows correctly placed city/town labels.
- Zoom in/out reveals/hides labels according to their zoom range.
- Theme toggle (light/dark/contrast) continues to work.

### LINKED ISSUE #3: tools/maps: add 'inspect' CLI and --preview-zoom flag for archive verification
## Problem

There is no fast way to answer "is this `.tdmap` sensible?" without flashing a device or opening `viewer.html`. When iterating on dithering, palette, or label extraction, this slows things down a lot.

## Proposal

Two small additions to `tools/maps/`:

### 1. `python -m tools.maps.archive inspect file.tdmap`

Prints:
- Format version and header fields
- Bounds (once issue for header metadata lands)
- Tile count per zoom level
- Label count by type (city/town/village/suburb/road/water)
- First N labels per zoom
- Average + min + max compression ratio per zoom
- Palette swatches as ANSI color blocks

### 2. `pmtiles_to_tdmap.py --preview-zoom N --preview-out preview.png`

Render a single zoom level as a flat PNG mosaic to inspect dithering/palette choices before committing to a multi-zoom build. Should be fast (minutes, not hours).

## Acceptance

- `inspect` runs in under 2 seconds on a ~100MB archive.
- `--preview-zoom` produces a correctly laid-out PNG for the chosen zoom.

### LINKED ISSUE #4: tools/maps: incremental rebuild via source-tile hashing
## Problem

`pmtiles_to_tdmap.py` has checkpoint resume every 500 tiles, but no change detection. Tweaking palette, dithering, or road priority requires a full re-render even though most tiles are unchanged.

## Proposal

- Store a hash of `(source_mvt_bytes + render_config_fingerprint)` per tile in a sidecar index (e.g. `.tdmap.cache.json`).
- On rebuild, compare hashes and skip tiles whose hash is unchanged — copy their compressed bytes directly from the previous archive.
- `render_config_fingerprint` = hash of palette + dithering params + feature priority. Bump it deliberately when behavior intentionally changes.
- Add a `--no-cache` flag to force a full rebuild.

## Acceptance

- Re-running the build with no source/config changes finishes in under 10% of the original time.
- Changing the palette alone triggers a full re-render (correct invalidation).
- Output archive bytes are identical to a from-scratch build.

### LINKED ISSUE #5: tools/maps: regression fixture for land/road coordinate desync
## Problem

Commit `48ad897` fixed a nasty bug where the land mask was drawn with Y-up geographic coordinates while MVT road layers used Y-down screen coordinates, so roads floated over the wrong land on certain tiles. The fix is purely in code — there is no regression test, so it is easy to reintroduce the next time rendering is refactored.

## Proposal

- Commit a tiny fixture: a hand-picked coastal MVT tile + the expected dithered PNG.
- Add a pytest (or plain script) under `tools/maps/tests/` that renders the fixture and compares against the golden image with a small tolerance (hamming distance or byte equality if determinism holds).
- Wire it into CI for the `tools/maps/` path.

## Acceptance

- Test exists, passes on `main`.
- Deliberately reverting `48ad897` causes the test to fail.

### LINKED ISSUE #6: tools/maps: ship pre-quantized land mask sidecar to avoid 24MB download
## Problem

`tools/maps/land_mask.py` downloads ~24MB of Natural Earth / OSM simplified land polygons on first run. This is slow in CI, flaky on poor connections, and blocks reproducible builds.

## Proposal

- Pre-quantize the mask into a compact per-zoom raster bitset (one bit per tile, "is this tile mostly land?") and commit it (or attach as a GitHub release asset).
- `land_mask.py` loads the bitset first; only falls back to downloading + quantizing the polygons when the user explicitly requests a non-default resolution.
- Expected size: a single global bitset at z0-z8 is well under 1MB.

## Acceptance

- First-time `pmtiles_to_tdmap.py` runs in CI without network access to OSM.
- Existing `is_land_tile` / `get_land_fraction` / `tile_intersects_land` behavior is unchanged at default resolution.

### LINKED ISSUE #7: tools/maps: split pmtiles_to_tdmap.py into render / labels / archive stages
## Problem

`tools/maps/pmtiles_to_tdmap.py` is 1,210 lines and mixes three concerns: MVT-to-raster rendering, label extraction, and TDMAP archive writing. This makes each stage hard to test in isolation and bloats the blast radius of any change.

## Proposal

Refactor into three modules, keeping `pmtiles_to_tdmap.py` as the CLI glue:

- `tools/maps/render.py` — MVT → dithered 3bpp raster. Pure function of (MVT bytes, zoom, palette, dithering params) → bytes.
- `tools/maps/labels.py` — MVT → list of `{lat, lon, text, type, zoom_min, zoom_max}` entries. Pure function.
- `tools/maps/archive.py` — already exists; keep as the writer/reader.

Each module gets a small unit-test file. The CLI orchestrates + handles parallelism + checkpoints.

## Acceptance

- No behavior change: output archives byte-identical to the pre-refactor build for a reference input.
- Each stage is testable without spinning up the full pipeline.

### LINKED ISSUE #8: tools/maps: enrich TDMAP header with bounds, region name, source hash, build timestamp
## Problem

The current 33-byte TDMAP header records format/version/compression/counts/offsets but has no provenance. Two `world.tdmap` files on disk are indistinguishable without opening them, and the device cannot tell the user what region is loaded.

## Proposal

Bump to TDMAP v5 with an extensible metadata block immediately after the fixed header:

| Field            | Size     | Notes                                |
|------------------|----------|--------------------------------------|
| region_name      | varlen   | UTF-8, e.g. "Netherlands"            |
| bounds           | 16 bytes | min_lat_e6, min_lon_e6, max_lat_e6, max_lon_e6 |
| source_hash      | 32 bytes | SHA-256 of input PMTiles             |
| build_timestamp  | 8 bytes  | UNIX epoch seconds                   |
| tool_version     | varlen   | `pmtiles_to_tdmap.py` version string |

Keep reader backwards-compatible with v4 archives.

## Acceptance

- `inspect` CLI (#3) can surface all fields.
- Device UI can show "Netherlands, built 2026-04-01" rather than a bare filename.
- v4 archives still load.

### LINKED ISSUE #9: tools/simulator: add TDMAP loader mock for browser iteration
## Problem

There is no way to iterate on a map screen in the browser simulator without flashing a device. Once a new map screen lands (#10, #11), this will be the fastest feedback loop for UX tweaks.

## Proposal

- Add `tools/simulator/mock/map_archive.js` that reads a `.tdmap` blob (drag-drop or fetched URL) and exposes the same API as the forthcoming `lua/services/map_archive.lua`.
- Reuse the RLE / 3bpp decode logic already present in `tools/maps/viewer.html` — extract to a shared JS module so both tools decode identically.
- Hook into the existing `ez.storage.read_bytes` mock so `require('services.map_archive')` works unmodified.

## Acceptance

- Loading a `.tdmap` in the simulator and running `ui.push_screen('$screens/map.lua')` renders tiles + labels correctly.
- No code-path divergence from the device reader.

### LINKED ISSUE #10: lua/services/map_archive.lua: async v4 TDMAP reader with LRU cache and parent-tile fallback
## Problem

The old map viewer (pre-`f200c36` restructure) embedded archive I/O, LRU caching, and label parsing directly inside the screen. With the new ezui framework, the data layer should be extracted so widgets are thin and the reader can be reused (location picker, share-location, breadcrumb trails).

## Proposal

New module `lua/services/map_archive.lua`:

- `open(path) -> archive` — reads header, palette, tile index; validates magic/version; returns handle or error.
- `get_tile(archive, z, x, y) -> tile_data | :pending | nil` — non-blocking. On miss, schedules an async load via `ez.async.spawn` and returns `:pending`. Hits return decoded 3bpp bytes immediately.
- `get_parent_fallback(archive, z, x, y) -> (parent_tile, src_x, src_y, src_w, src_h)` — walk up zoom levels until a cached ancestor is found, return coordinates suitable for `ez.display.draw_indexed_bitmap_scaled`.
- `labels_in_bounds(archive, z, lat0, lon0, lat1, lon1) -> list` — filtered by viewport and zoom range.
- LRU cache: 16 tiles ≈ 384KB PSRAM (matches the old viewer's tuning).
- Negative-cache of known-missing `(z,x,y)` triples; cleared on zoom change.
- Parses v4 labels (and v5 when #8 lands).

## Acceptance

- Unit-testable without a screen attached.
- Simulator TDMAP mock (#9) can plug in transparently.
- No blocking SD reads from the main render loop.

### LINKED ISSUE #11: lua/ezui/widgets/map_view.lua: reusable map widget node
## Problem

ezui has no map-style widget (grid panning, zoom, tile rendering, label overlay). The old shell-based map viewer was a top-level screen that couldn't be embedded elsewhere.

## Proposal

Register a new node type `map_view` via `node_mod.register(...)` following the `color_swatch` pattern in `lua/screens/settings.lua:9-49`.

Props:
- `archive` (handle from `map_archive.open`)
- `center_lat`, `center_lon`, `zoom`
- `show_labels` (bool)
- `overlay_fn(d, x, y, w, h, project)` — optional draw hook for GPS dot, pins, route lines
- `on_move(lat, lon, zoom)` — emitted when the user pans/zooms

Behavior:
- `focusable = true`; owns the trackball while focused.
- `draw`: compute visible tile range, call `ez.display.draw_indexed_bitmap` per tile, use `_scaled` with parent fallback when a tile is pending.
- `on_key`: trackball directions pan at sub-tile precision; `+`/`-` zoom; `L` toggle labels; `T` toggle theme.
- Reads palette from `ezui.theme` — **no** hardcoded RGB565 tables.
- Exposes a `project(lat, lon) -> screen_x, screen_y` helper so overlays stay aligned.

## Acceptance

- Can be embedded inside any ezui layout (vbox/hbox/zstack).
- Works identically on device and in the simulator (once #9 lands).
- Blocks #11.

### LINKED ISSUE #12: lua/screens/map.lua: new map screen with GPS integration
## Problem

The map screen was dropped when the Lua tree was restructured (`f200c36`) and never ported to ezui. Users have no way to view offline maps in the current build.

## Proposal

New screen `lua/screens/map.lua`, built on top of `services/map_archive` (#10) and `ezui.widgets.map_view` (#11).

Layout:
```
title_bar("Map", { back = true, actions = { theme, labels, gps } })
map_view { ... }                                   -- flex: 1
hbox { lat_lon_text, zoom_text, gps_indicator }    -- 16px status strip
```

Behavior:
- `on_enter`: open archive, show spinner until ready, restore last view from `ez.storage.get_pref("map_last_view")`.
- Subscribe to `ez.gps` updates via the bus; draw position dot + accuracy circle via the `map_view` overlay hook.
- `G` short-press: jump once to current GPS fix. Long-press: follow-mode (recenter on every update).
- `on_leave`: persist center/zoom so reopening is instant.
- Shows the archive's region/name/build-date (once #8 lands) in the title bar or an info overlay.

## Acceptance

- Navigable end-to-end on device and simulator.
- GPS dot tracks correctly as you pan; stays aligned when zooming.
- Battery-friendly: no redraws when idle.

### LINKED ISSUE #13: docs: update README for ezui framework and reinforce early-development stage
## Problem

The README is out of date in several places and understates how early the project is. New users land on it expecting stability and a documented path that no longer exists.

### Stale content

- **Project Structure** (lines ~124-140) still shows `data/scripts/boot.lua` and `data/scripts/ui/screens/`. Since commit `f200c36` the tree lives under `/lua/` and is embedded as bytecode into the firmware.
- **Getting started** (line ~65) says *"Edit scripts directly on the SD card in `/data/scripts/`"*. This no longer works — scripts are compiled in; users must rebuild + flash.
- **UI System example** (lines ~205-222) shows the old shell API (`render(display)`, `handle_key`). The current framework is declarative ezui: screens expose `build(state)` returning a node tree. See `lua/screens/settings.lua` for the real pattern.
- **Remote Control** example references `Debug.memory()` which no longer exists (use `collectgarbage('count')`).
- No mention of `ezui` anywhere, despite it being the entire UI layer.
- No mention of the map viewer being temporarily absent (tracked in #10/#11/#12).

### Missing: early-dev stage messaging

The only warning is a single line on the flash section. For a project in active churn (Lua tree restructure, UI framework rewrite, TDMAP format bumps, binding renames), the README should make the alpha/early-dev state obvious on first screen.

## Proposal

- Add a prominent banner near the top: **"Early development — expect breaking changes between releases. Not yet suitable for production use."**
- Rewrite *Project Structure* to reflect the current tree.
- Replace the *Getting started* SD-card instruction with the rebuild-and-flash workflow (or link to the simulator for faster iteration).
- Replace the *UI System* example with a real ezui `build(state)` screen.
- Add a short *UI Framework* section describing ezui (declarative nodes, focus, theme) and linking to `lua/ezui/`.
- Fix the `Debug.memory()` reference.
- Note that offline maps are temporarily not wired into the UI (link to tracking issues).

## Acceptance

- A new user reading only the README gets an accurate picture of where the code lives, how to run it, and how stable it is.

### LINKED ISSUE #17: Map tiles: road lines don't stitch across tile boundaries
## Symptom

At z10 panning over the Netherlands, horizontal and vertical seams are visible between adjacent map tiles. **Polygon fills (coastlines, parks, water bodies) stitch cleanly** — a land edge crossing a tile boundary looks continuous. But **road lines break at every tile boundary**: a motorway entering tile (x, y) at its south edge doesn't visually connect to the same motorway leaving tile (x, y+1) at its north edge.

The effect is most obvious on long straight arterials that span many tiles (A1, A2, etc.) — they appear as a stack of disconnected road segments rather than one continuous line.

## Root cause

`render_vector_tile` in `tools/maps/render.py` (ex-`pmtiles_to_tdmap.py`) renders each tile in isolation from its own MVT data. Two interacting factors cause the break on lines but not on polygons:

1. **MVT geometry is clipped at the tile extent.** A road that crosses from tile A to tile B has two endpoints in A's MVT (entry point plus snapped endpoint at the extent edge) and two in B's (snapped start plus exit point). Polygon fills also get clipped this way, but PIL's `draw.polygon` fills up to and including the clipped edge pixels, so adjacent tiles' fills touch.

2. **PIL's `draw.line(..., width=W)` draws a butt cap at each endpoint.** When the endpoint lands on or slightly past the tile edge and the stroke is thin (width ≤ 2), the cap renders at a different sub-pixel position on each side of the seam. Integer rasterization then snaps the two caps to pixel coordinates that disagree by 1–2 pixels, leaving a visible gap.

Combined, polygons tolerate the clip, lines don't.

## Evidence

- Verified by rendering tiles `z10/527/334` and `z10/527/335` individually and stitching in Python: coastline continuous, roads have visible breaks at y=256.
- Confirmed on device against `netherlands-zlib.tdmap` at z10 center (52.5488, 5.25).

## Proposed fixes (ranked)

1. **Render with an overlap halo.** Temporarily bump the rendering extent to e.g. `TILE_SIZE + 32` pixels per side, draw all features that fall in the expanded bbox, then crop the centre 256×256. Road endpoints now land inside the expanded area instead of on the boundary, so their caps are drawn before clipping removes them. Cleanest fix; costs ~25% more CPU per tile at build time, no format change, no runtime cost.

2. **Minimum line width 2.** Bump every entry in `ROAD_STYLE` to at least 2 px. Much cheaper, but makes z≥13 render look heavy and doesn't actually close the gap — the two caps still disagree, just more visibly.

3. **Render neighbour-tile roads into the current tile.** For each tile, also pull the four neighbours' MVT data and draw any line whose clipped extent touches this tile's bbox. Correct, but 5× the MVT decodes and stitches its own caching problem.

4. **Build-time post-processing.** After rendering adjacent tiles, scan both sides of each shared edge and dilate road pixels to fill 1-pixel gaps. Lossy and brittle.

## Acceptance

- Long straight roads spanning ≥ 3 tiles render as a single continuous stroke at all zooms 8–14.
- No regression to polygon rendering, label placement, or file size.
- The regression fixture (`tools/maps/tests/test_land_road_alignment.py`) still passes (will need `--regen` since pixel-level output will shift slightly).

bvdaakster:
Two stacked chapters of work on the ezui shell.

### Status bar + back key
- Global status bar pinned at the top with battery, wifi, gps, clock, and an async-activity spinner. Screens opt out with `fullscreen = true` (games, pixel_fix).
- Per-screen title bars collapse into a compact sub-bar with a drawn left-arrow glyph + "Back" label. The T-Deck's physical back-arrow key (`BACKSPACE`) replaces Q as the default exit across shell screens.

### Typography
- Anti-aliased Inter at 11/13/17 px pre-rasterised by `tools/gen_aa_font.py` into `src/fonts/InterAA*.h`; new alpha-blend renderer in `src/hardware/aa_font.{h,cpp}`. Every shell widget renders through `small_aa`/`medium_aa`/`large_aa`.

### Desktop icons (PSP/PS3 flavour)
- Layered composition in `lua/screens/desktop.lua`: colour plate (`fill_round_rect`) + white glyph PNG + shared glass shim PNG + pre-blurred focus halo.
- Contacts is red, Map is green.
- Focused icon pulsates via a sine-wave `brighten_rgb565` factor on the plate colour.
- `tools/gen_icons.py` splits the old baked-in icons into plain white glyph PNGs plus a shared shim/glow, keyed by an RGB565 tint per entry.

### Framework fixes
- `_rebuild` persists widget-internal state (scroll offset, dropdown open/cursor, text-input cursor) across `set_state` so periodic screens (GPS live status) don't reset the user's interaction.
- Dropdown grabs `focus.enter_edit()` while open, handles ENTER in `on_key` (bypasses the editing-swallow path), and routes properly through `focus.handle_key`.
- Focus auto-scroll snaps to 0 when the first item is focused so headers above the first focusable become visible.
- Colour-swatch row consumes UP/DOWN to jump past the whole strip instead of stepping colour-by-colour.
- Scroll containers draw top/bottom edge shadows (`lua/ezui/shadows.lua`, generated by `tools/gen_shadows.py`) to hint overflow.
- W hotkey on desktop debounced against the known keyboard ghost-repeat.

### Audio (SND01 / snd.dev)
- Integrated the SND01 "Sine" pack. Attribution in the About screen and `data/sounds/snd01/CREDITS.md` (Yasuhiro Tsuchiya / Dentsu, https://snd.dev).
- Non-blocking `ez.audio.play_preloaded_async` binding; preload cap bumped to 32 so every sample stays resident in PSRAM.
- `services/ui_sounds.lua` preloads at boot and plays variants round-robin (5 `tap`, 5 `swipe`, 5 `type`). Widgets fire events on button/list-item activate, toggle change, focus change, text input character, dropdown confirm, and screen push/pop. Gated on a user pref.

### Settings refactor
- Top-level settings is now a category picker → Display / Keyboard / GPS / Sound / About. Each submenu has its own scrollable page.
- Sound page: UI-sounds toggle + master volume.
- About page: version + third-party credits (SND01, Lucide, Inter, MeshCore, LovyanGFX).

### Wallpapers
- `tools/gen_wallpapers.py`: cover-fits to 320×240, Floyd–Steinberg dithers into RGB565, re-encodes JPEG at q=82. Fixed a sign-extension bug in the dither loop. Regenerated `wp01..wp30`.

### Developer
- Main menu gains a **Developer** section with a **Widget kitchen sink** screen — every widget on one page for manual smoke-testing.

---

### Offline maps

- **New Lua map stack** — `services/map_archive`, `ezui/widgets/map_view`, `screens/map`. Async tile cache, parent-tile fallback, center crosshair, theme-aware labels with halo.
- **TDMAP v5** — adds a TLV metadata block (region name, bounds, source hash, build timestamp, tool version). Reader accepts both v4 and v5; writer emits v5 by default.
- **zlib compression** — new `ez.compression.inflate` binding using ESP32 ROM miniz. Amsterdam archive: 2.7 MB → 994 KB. World archive estimate: 478 MB → ~80 MB.
- **Better tools** — `inspect` CLI, `--preview-zoom` flag, `--region-name`, render-fingerprint-aware checkpoint that invalidates on config/source change, regression fixture for the coastal-tile desync bug.
- **Committable land mask sidecar** — 22 KB bit-packed raster replaces the 24 MB one-time download for fresh clones and CI.
- **Simulator catch-up** — drag-drop archive upload, pako-backed zlib mock, in-memory file store.

### GPS controls and overlay

- GPS settings screen under Settings → GPS. Power toggle, clock-sync policy (Never / At boot / Hourly), live fix + satellite readout, "sync clock now" action.
- Background sync coroutine in `services/gps`, started at boot.
- User position marker on the map: in-view target-style dot, off-view triangular arrow at the viewport edge pointing to the fix.
- `H` key — home-to-GPS (one-shot recenter without toggling follow-mode; `g` still toggles follow).

### Road seam fix

- `render_vector_tile` paints onto a `TILE_SIZE + 2*HALO` canvas with an `offset=HALO`, then crops to the visible centre. Closes #17.

### File size comparison (measured)

| Archive             | RLE      | zlib(9)  |
|---------------------|----------|----------|
| World (z0–z14)      | 478 MB   | ~80 MB   |
| Amsterdam (z10–z13) | 2.7 MB   | 994 KB   |

---

## Closes

- Closes #2 — viewer.html v4 label support
- Closes #3 — archive inspect CLI + `--preview-zoom`
- Closes #4 — regression fixture for land/road coordinate desync
- Closes #5 — pre-quantized land mask sidecar
- Closes #6 — incremental rebuild via source-tile fingerprint
- Closes #7 — split `pmtiles_to_tdmap` into render/labels/archive
- Closes #8 — TDMAP v5 header with bounds / region / hash / timestamp
- Closes #9 — TDMAP loader mock for the simulator
- Closes #10 — `lua/services/map_archive`
- Closes #11 — ezui `map_view` widget
- Closes #12 — `lua/screens/map`
- Closes #13 — README update for ezui + early-dev banner
- Closes #17 — road seam artifact at tile boundaries

Follow-ups tracked separately: #14 (deprecate RLE in a future format bump), #15 (coroutine-aware async driver for the Wasmoon simulator).

## Test plan

Map
- [x] `pio run -t upload` succeeds
- [x] Map screen pushes cleanly; tiles render; labels readable; status strip shows coords + zoom; crosshair anchors the zoom pivot
- [x] +/- zoom and trackball pan work; `l` toggles labels; `g` jumps/follows GPS; `t` cycles theme; `h` homes to GPS
- [x] User-position marker in-view / off-screen arrow at viewport edge
- [x] `pytest tools/maps/tests/` passes after the render-halo change (<0.5% drift)
- [x] Stitched adjacent z10 tiles visibly improved — roads continue across y=256
- [x] zlib round-trips through `ez.compression.inflate` on-device
- [x] v4 archives still open alongside v5

Shell
- [x] Global status bar renders on every shell screen; spinner activates on async work
- [x] Desktop icons render colour plates with shim/glow; focused icon pulsates
- [x] AA Inter fonts render throughout shell widgets (button, toggle, slider, dropdown, text_input, list_item, title_bar, status_bar, desktop labels)
- [x] BACKSPACE pops shell screens; drawn back arrow visible in sub-bar
- [x] GPS dropdown survives the 1 Hz live-status rebuild; "Hourly" selectable
- [x] Scrolling into the first item snaps the viewport to 0
- [x] SND01 samples audible (tap on focus change, button on desktop-icon activate, toggle_on/off on toggle, type on text input, transition_up/down on push/pop)
- [x] UI-sounds toggle + master volume work; About page credits populate


bvdaakster:
@claude review

Issue #71, opened 2026-05-10
Expand notifications service integration points


bvdaakster:
Today the notifications service (`lua/services/notifications.lua`) is only wired up to one source: OTA progress in `lua/boot.lua:217`. The infrastructure (queue, toast overlay, sticky/action support, source-based dedupe via `dismiss_source`) is general-purpose and could surface a lot of events the user currently misses.

Proposing the following integration points. Each follows the same pattern as the OTA hook -- subscribe to an existing bus topic in `boot.lua` (or in the originating service) and translate into `notifications.post{...}`.

## High-value (user-visible events the user might miss)

- **Direct messages** -- subscribe to `dm/message` from `services/direct_messages.lua`. Action: open the conversation. Suppress when the chat screen for that contact is already focused.
- **DM delivery failures** -- `dm/status` with status=`unconfirmed` after retries exhausted.
- **Channel mentions** -- `channel/message` filtered to mentions of the user's node name, or per-channel "notify" opt-in pref. Posting on every public message would be noisy.
- **File transfer offers** -- `file/offer` from `services/file_transfer.lua`. Action: open the transfer screen.
- **File transfer complete / failed** -- `file/done` ("Received foo.txt", action opens via `services.apps`) and `file/error`.
- **Inbound share cards via custom packets** -- contact share / channel invite / time share arriving while the user is elsewhere. Today they only land silently in chat history; surface them with the same actions the bubble context menu offers (Add contact / Join channel / Sync clock).

## System health

- **Radio init failed** (the `!RF` indicator) -- sticky notification at boot if the LoRa module didn't come up. Currently the only signal is a tiny corner indicator that's easy to miss.
- **Low battery** -- threshold crossings at 20% / 10% / 5% via `ez.system.get_battery_percent()`. Use a single `source = "battery"` and `dismiss_source` so a higher level dismisses the previous one.
- **Brownout / panic recovery** -- on cold boot, check `esp_reset_reason()` and post once if the previous boot ended in panic/brownout, with a "View log" action that opens `log_persist`.
- **SD card removed** -- `services/file_transfer`, the map loader, and `log_persist` all silently degrade when the SD wrapper desyncs (see #51). A notification when SD goes away would explain why maps stop loading.
- **SD card connected** -- transition from "no SD" to "SD mounted" while running. Useful confirmation after the user inserts a card mid-session, and pairs naturally with the "SD removed" event.

## Background services completing

[...]

PR #80, opened 2026-05-10
feat(channels): per-channel history limit + notification mode

### LINKED ISSUE #72: Per-channel settings: history retention + notification mode
Channels currently have no per-channel configuration beyond join/leave + hide. Two settings would make the chat experience a lot more tunable, and they pair naturally because they're both per-channel preferences exposed from the same settings sheet.

## 1. Per-channel history retention

Today `lua/services/channels.lua` keeps a hard-coded ring buffer of `MAX_HISTORY = 50` messages per channel (`lua/services/channels.lua:7`), in RAM only -- nothing is persisted across reboots. That's fine for `#Public` but limiting for low-traffic channels where the user wants to scroll back.

Proposal: per-channel `history_limit` setting, with options like:

- **None** (don't keep history -- messages disappear once scrolled off / on reboot)
- **50 messages** (current default)
- **200 messages**
- **500 messages**
- **1000 messages**

The trim loop at `lua/services/channels.lua:94` reads the per-channel limit instead of the constant. Persistence to LittleFS would be a follow-up (the existing `PREF_KEY = "joined_channels"` format at `lua/services/channels.lua:8` would need to grow to carry the limit, or move to a JSON sidecar -- worth a separate decision).

Settings UI: add an "M-key" (Alt+M) action on the channel list / channel screen that opens a per-channel settings sheet.

## 2. Per-channel notification mode

Companion to #71 (notifications integration). Each channel gets one of:

- **All messages** -- post a notification on every `channel/message` for this channel (action = open the channel screen).
- **Mentions only** -- only when the message text contains the user's node name (case-insensitive, word-boundary). This should be the default for `#Public` so the user isn't drowned.
- **None** -- never post notifications for this channel.

Suppress posts when the channel screen for that channel is already focused (use the `post_unless_focused` helper proposed in #71).

Storage: per-channel pref key alongside the history limit. Read in the `boot.lua` subscriber that translates `channel/message` into `notifications.post`.

## Settings UI sketch

Per-channel settings sheet (reachable via Alt+M on the channel row or inside the channel screen):

```
Channel: #Public
--------------------------
History         [ 200 messages > ]
Notifications   [ Mentions only > ]
Hide channel    [ off ]
Leave channel
```

Defaults:

- `#Public`: 50 messages history, "Mentions only" notifications.
- Other joined channels: 200 messages history, "All messages" notifications.

## Out of scope (call out for a follow-up)

- Persisting message history across reboots is a bigger lift (storage format, size budget, write amplification on flash). This issue covers the in-RAM retention dial; a separate issue should track on-disk persistence if/when wanted.
- DM equivalents -- DMs deserve the same treatment, but the storage shape is different. Track separately.

Depends on / overlaps with #71 for the notification post path.

xtr-bas:
## Summary

Closes #72.

`services/channels` gains two pref-backed dials per channel:

- **history_limit**: 0 / 50 / 200 / 500 / 1000. `store_message` reads it instead of the previous hard-coded `MAX_HISTORY=50`, and `set_history_limit` trims live history immediately. Defaults: 50 for `#Public`, 200 for everything else.
- **notify_mode**: `all` / `mentions` / `none`. Read by the boot-side `channel/message` subscriber that posts into `services.notifications`. Defaults: `mentions` for `#Public`, `all` elsewhere.

NVS caps key length at 15 chars, so the per-channel pref keys use short prefixes (`ch_h`, `ch_n`) and truncate the sanitized channel name to fit. Without that the dial appears stuck on its default because `set_pref` silently fails.

UI: new `screens/chat/channel_settings.lua` (history dropdown, notify dropdown, hide toggle, Leave button). Reachable from the channel chat's Alt+M menu alongside the existing Share Time action.

Notification wiring in `boot.lua` honours the per-channel mode and suppresses when the channel chat for that channel is already focused. The focus check is inlined here; a follow-up will route it through `services.notifications.post_unless_focused` once that helper lands from #71.

## Depends on

#71 contributes the `post_unless_focused` helper. This PR is independent (inlines the equivalent check) so it can ship before or after #71. Cleanup tracked separately.

## Test plan

- [x] Build + flash, defaults: `#Public` -> 50 / mentions; other -> 200 / all.
- [x] `set_history_limit` / `set_notify_mode` round-trip through NVS.
- [x] `channel/message` for an unset channel with default `all` -> notification.
- [x] `channel/message` on `#Public` (mentions): non-mention text suppressed; text containing the user's node name posts.
- [x] `channel/message` on a channel set to `none` -> always suppressed.
- [x] `channel/message` while the chat for that channel is focused -> suppressed.
- [x] Channel settings screen renders correctly via `--text` capture.
- [x] Channel chat Alt+M menu now has "Channel settings" + "Share time".

### DIFF
--- a/lua/boot.lua
+++ b/lua/boot.lua
@@ -416,6 +416,53 @@ local function boot_sequence()
         end
     end
 
+    -- ---- Channel messages ----
+    -- Honour the per-channel notify mode (none / mentions / all).
+    -- "Mentions" matches the user's node name as a case-insensitive
+    -- substring of the message text -- word-boundary matching would
+    -- be more accurate but Lua's patterns don't have a real \b and
+    -- the false-positive rate on a substring match is low for the
+    -- typical short device name.
+    local channels_svc = require("services.channels")
+    ez.bus.subscribe("channel/message", function(_topic, msg)
+        if type(msg) ~= "table" or msg.is_self then return end
+        local name = msg.channel
+        if not name then return end
+        local mode = channels_svc.get_notify_mode(name)
+        if mode == "none" then return end
+        if mode == "mentions" then
+            local my = ez.mesh and ez.mesh.get_node_name and ez.mesh.get_node_name()
+            if not my or my == "" then return end
+            local hit = msg.text and msg.text:lower():find(my:lower(), 1, true)
+            if not hit then return end
+        end
+        notifications.post_unless_focused({
+            title  = name,
+            body   = string.format("%s: %s",
+                                   msg.sender_name or "?",
+                                   (msg.text or ""):sub(1, 80)),
+            source = "channel",
+            action = {
+                label    = "Open",
+                on_press = function()
+                    local screen = require("ezui.screen")
+                    local Chat   = require("screens.chat.channel_chat")
+                    screen.push(screen.create(Chat, { channel = name }))
+                end,
+            },
+        }, function(inst)
+            -- Suppress only when the channel-chat screen for this
+            -- exact channel is on top. Gate on the screen *type*,
+            -- not just `_state.channel`, because the channel
+            -- settings sheet also carries the channel name in its
+            -- state and would otherwise eat notifications.
+            local Chat = require("screens.chat.channel_chat")
+            return inst._def == Chat
+                   and inst._state
+                   and inst._state.channel == name
+        end)
+    end)
+
     -- Apps registry: file-type → handler for the file manager. Built-in
     -- handlers register themselves here; screens opened from the registry
     -- are loaded lazily on first `open()` so unused apps don't pull their
--- a/lua/screens/chat/channel_chat.lua
+++ b/lua/screens/chat/channel_chat.lua
@@ -208,6 +208,16 @@ function ChannelChat:menu()
     local channel = self._state.channel or "#Public"
     return {
         {
+            title = "Channel settings",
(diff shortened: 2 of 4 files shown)


xtr-bas:
Rebased onto current `origin/test` (which now includes #78 and #79's `notifications.post_unless_focused` helper) and folded both review comments into the same commit:

- **`store_message` trim with `limit=0`**: `while #h > math.max(limit, 1)` so the just-added message survives at least the current frame; the explicit `set_history_limit` setter still drains all the way to 0 since that's a deliberate user action on pre-existing history.
- **Focus suppression match**: predicate now checks `inst._def == ChannelChat` in addition to `_state.channel == name`. Channel Settings (which also stores the channel in `_state.channel`) no longer eats notifications.

While here, switched the `channel/message` subscriber from the inline focus check to `notifications.post_unless_focused` (the helper from #79 is now available since this PR rebased onto it). That **closes #81**.

### Verified on-device
- `#Public` set to `history_limit=0`, sent a message via `channels.send`, history kept it (`#h = 1`). (Sorry — that send actually transmitted on `#Public`. I should have used a non-joined channel; won't repeat.)
- Channel Settings on top + simulated `channel/message` for `#Public` -> notification posted (regression case, was previously suppressed).
- ChannelChat on top + simulated `channel/message` -> notification suppressed (positive case, still correctly type-gated).

PR #91, opened 2026-05-11
docs(manual): regenerate menu + settings pages from current sources

bvdaakster:
## Summary

The on-site Menu Reference (`/manual/menu/`) was empty (the screen
table only had its header row) because the generator still pointed at
`data/scripts/ui/screens/main_menu.lua`, which doesn't exist any
more. The Settings Reference was empty for the same reason --
`settings_category.lua` was replaced by `lua/services/prefs_registry.lua`.

Both auto-generated pages now reflect the current sources.

## Changes

- **`tools/generate_lua_docs.py` — menu parser** (commit 1):
  - Looks up `lua/screens/menu.lua` and walks the new tabbed
    `CATEGORIES = {...}` structure (was the flat `items = {...}` with
    per-item `shortcut`).
  - `MenuItem` no longer has `shortcut` / `enabled`; new
    `MenuCategory` groups entries.
  - New brace-balancer skips Lua line comments (`--`) and long
    comments (`--[[ ... ]]`) — quotes inside the file's header
    docstring used to throw off the old balancer, which is why it
    silently returned an empty match.
  - Generated page renders one markdown table per tab and lists
    navigation that matches the T-Deck Plus keyboard (Tab/Alt+M to
    open the menu, Left/Right for tabs, Backspace for back — there
    is no Esc or Ctrl key).

- **`tools/generate_lua_docs.py` — settings parser** (commit 2):
  - Reads `lua/services/prefs_registry.lua` (was a non-existent
    `settings_category.lua`).
  - Treats `-- ---- Section ----` comment headers as category
    breaks; every `{ key = ..., type = ..., default = ..., ... }`
    becomes a row.
  - Compact per-category table (NVS key / Type / Default / Range
    or Options / Description) replaces the old per-setting H3
    sections, which depended on editor-only schema fields (`label`,
    `suffix`, `toggle`/`option`) that no longer exist in the
    registry.

## Numbers
- Menu page: **0 → 41 entries across 7 tabs**
- Settings page: **0 → 23 prefs across 7 categories**

## Test plan

- [x] `python tools/generate_lua_docs.py` reports the expected
      counts and writes both pages without errors.
- [x] Generated `docs/manual/menu/index.md` lists every screen reachable
      from `lua/screens/menu.lua`.
- [x] Generated `docs/manual/settings/index.md` lists every entry from
      `lua/services/prefs_registry.lua`.
- [x] Manual index (`docs/manual/index.md`) shows the correct totals in
      the contents table and inlines a per-tab summary of the menu.

### DIFF
--- a/tools/generate_lua_docs.py
+++ b/tools/generate_lua_docs.py
@@ -2243,7 +2243,7 @@ def generate_shell_guide_index(categories: List[SettingsCategory], modules: Dict
         "- **Arrow keys** or **Trackball**: Move focus through the list",
         "- **Enter**: Open the focused item",
         "- **Backspace** (back-arrow key): Go back to the previous screen",
-        "- **Tab** (Alt+M): Open the main menu from the desktop",
+        "- **Alt+M**: Open the main menu from the desktop",
         "- **Left** / **Right**: Switch between menu tabs",
         "",
         "## Contents",
@@ -2407,7 +2407,7 @@ def generate_menu_reference(menu_categories: List[MenuCategory]) -> str:
     lines = [
         "# Menu Reference",
         "",
-        "Press **Tab** (or Alt+M) on the desktop to open the main menu. The menu is",
+        "Press **Alt+M** on the desktop to open the main menu. The menu is",
         "organised into a horizontal tab strip; each tab lists the screens it owns.",
         "",
         "## Navigation",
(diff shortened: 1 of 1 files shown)


bvdaakster:
@claude fix the feedback

PR #99, opened 2026-05-11
docs(autopilot): fan out per-issue work across subagents

bvdaakster:
## Summary

- Reshapes `/autopilot` from a sequential per-issue loop into an **orchestrator + workers** model.
- The skill itself now handles pre-flight, batch claimability checks, worktree creation, and host-tool staging (including copying gitignored `tools/bin/luac32` forward so fresh worktrees can `pio run`).
- One `general-purpose` subagent per surviving issue is spawned in a **single message**, so the runtime fans them out in parallel. The orchestrator is explicitly forbidden from editing code, building, or opening PRs itself.
- Replaces the "2 consecutive hard failures" stop condition (a sequential concept) with "all agents in the batch SKIPPED". Caps a run at one batch of five.
- Notes that direct push to `test` is blocked by branch protection, so the `.worktrees/` gitignore must land via a PR rather than block the batch.

## Test plan

- [x] Skill markdown still parses (frontmatter intact, headings well-formed).
- [ ] Next `/autopilot` run successfully fans out a batch of subagents in parallel from a single message (will validate end-to-end on next invocation).


xtr-bas:
Addressed all three review comments in `da78ca2`:

- **`.claude/skills/autopilot/SKILL.md:103`** -- hardcoded `/home/bastiaan/Desktop/tdeck-os/` replaced with `<abs-repo-root>`, captured at Setup via `git rev-parse --show-toplevel`. Note added above the template explaining the substitution requirement.
- **`.claude/skills/autopilot/SKILL.md:92`** -- added a pre-flight `ls tools/bin/luac32` check at the top of the per-worktree-prep section. Fails fast with a clear "run pio run in the main checkout first" hint instead of spawning five agents that all die on the missing toolchain. The per-issue copy step retains the assertion in comment form.
- **`.claude/skills/autopilot/SKILL.md:31`** -- Setup step 2 now spells out the full branch + commit + push + `gh pr create --base test` + `git checkout test` dance, so the main checkout ends Setup back on a clean `test`.

`aac3475` merges latest `origin/test` into the branch; `pio run` clean.

xtr-bas:
Addressed two new inline comments in \`ceed92d\`:

- **\`.claude/skills/autopilot/SKILL.md:122\`** -- \`git worktree add -b\` only seeds the branch from \`origin/test\`; it does not set an upstream. Reworded "tracking it" to "no upstream set yet -- first push uses \`git push -u\`".
- **\`.claude/skills/autopilot/SKILL.md:143-144\`** -- dropped the invalid "(1M context)" suffix from the \`Co-Authored-By\` trailer. The trailer is still hardcoded to Opus 4.7; making it match the executing subagent model dynamically is a separate concern and left for a follow-up.

\`6603920\` merges latest \`origin/test\`; \`pio run\` clean.

PR #103, opened 2026-05-13
feat(signal-test): scroll viewport + ezme.sh URI for SIGT DMs

xtr-bas:
## Summary
- Wrap the signal test screen below the title bar in a `ui.scroll` container so the Start/Stop button is reachable on the 240 px screen via trackball, UP/DOWN, or finger drag instead of being clipped under the chart.
- Replace the `[SIGT]P <nonce>` / `[SIGT]R <nonce>` DM text with a new sharing verb `https://ezme.sh/#sigt/v1?k=<P|R>&n=<nonce>`, matching how contact / channel-invite / time shares already ride the DM path.
- Add `sharing.is_protocol_message` and use it to filter SIGT pingpong DMs out of `dm.get_conversations` last-message previews, the DM thread view, unread counts, and DM toast notifications, so an active DM-mode test no longer reshuffles the chat list, spams toasts, or buries real chat under protocol carriers.

## Test plan
- [x] `pio run` builds clean.
- [x] On-device: open Signal Test, press DOWN — Start button auto-scrolls into view.
- [x] `sharing.encode_sigt("P", "abc1234")` round-trips through `sharing.parse` and `sharing.is_protocol_message` returns `true`; plain text and other share kinds (time) return `false`.
- [ ] Two devices: run a DM-mode test; confirm chat list preview and unread badge ignore the pingpong, then `p` purges the SIGT entries from history as before.

### DIFF
--- a/lua/boot.lua
+++ b/lua/boot.lua
@@ -244,6 +244,10 @@ local function boot_sequence()
     -- with that contact -- the screen itself shows the new bubble.
     ez.bus.subscribe("dm/message", function(_topic, msg)
         if type(msg) ~= "table" or msg.is_self then return end
+        -- Protocol-carrier DMs (signal-test pingpong) decrypt
+        -- successfully but aren't messages the user wrote, so they
+        -- must not raise a notification toast.
+        if require("services.sharing").is_protocol_message(msg) then return end
         local key  = msg.sender_key
         local name = msg.sender_name or "Unknown"
         local body = msg.text and msg.text:sub(1, 80) or nil
--- a/lua/screens/chat/dm_conversation.lua
+++ b/lua/screens/chat/dm_conversation.lua
@@ -369,7 +369,20 @@ function DMConversation:build(state)
     local msgs = dm_svc.get_history(key)
     local content_items = {}
 
-    if #msgs == 0 then
+    -- Hide signal-test pingpong DMs from the conversation view. They
+    -- still live in history (the tester's `p` shortcut purges them on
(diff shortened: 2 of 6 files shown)


xtr-bas:
Addressed both review comments in d92b953:

**1. Silent-send vector (`is_protocol_message` on raw text).**
Replaced the text scan with a stamped `msg.protocol` field. Stamping happens in `direct_messages` at the DM seam, gated on `signal_test.matches_protocol(msg)` -- which only returns true while the signal-test screen is open. A SIGT-shaped DM arriving outside an active test goes through the normal chat path (toast, unread badge, bubble), so a peer hand-typing the URL no longer disappears.

`dm.send` now accepts an `opts.protocol` so outbound SIGT pings/replies stamp themselves at the originator. The flag is persisted to disk so stamps survive reboots.

**2. Legacy `[SIGT]P/R <nonce>` entries stranded after upgrade.**
`signal_test.parse_dm` now accepts both the URL form and the legacy text form. `purge_dm_history` reuses `parse_dm`, so the `p` shortcut in the test screen cleans up either format. `matches_protocol` also walks both forms while a test is active, so legacy bubbles get classified back into protocol-hidden state when the operator re-opens the screen.

Smoke-tested on device:
- `is_protocol_message({ text = "https://ezme.sh/#sigt/v1?k=P&n=x" })` → `false` (no stamp)
- `is_protocol_message({ text = "anything", protocol = "sigt" })` → `true`
- `matches_protocol(url)` while active → `true`; while inactive → `false`
- `matches_protocol(legacy)` while active → `true`

PR #104, opened 2026-05-13
feat(input-lock): Shift+Alt+L / Shift+Alt+U global input lock

### LINKED ISSUE #77: feat: input lock via Shift+Alt+L (Shift+Alt+U to unlock)
## Summary

Add a global input lock that prevents any keyboard or touch input from reaching the UI until the user explicitly unlocks the device. Useful for putting the T-Deck in a pocket / bag without it eating keypresses or stray touches.

## Behavior

- **Lock chord:** `Shift + Alt + L` -- toggles the device into "input locked" mode from anywhere in the UI.
- **Unlock chord:** `Shift + Alt + U` -- exits input locked mode.
- While locked:
  - All keyboard input is swallowed *except* the unlock chord.
  - All touch input is swallowed (no taps, no drags reach screens or widgets).
  - The trackball is also swallowed.
  - A persistent on-screen hint is shown (e.g. an overlay/banner reading `Locked -- Shift+Alt+U to unlock`) so the user isn't left wondering why nothing responds.
  - Background services keep running (mesh, GPS, notifications, etc.) -- this is purely an *input* lock, not a sleep/standby mode.
- Lock state should *not* persist across reboots (avoid bricking yourself with a stuck lock if the chord path regresses). On boot, always start unlocked.

## Implementation notes

- **`CLAUDE.md` correction needed:** the "Keyboard layout (T-Deck Plus)" section currently claims there is no dedicated Shift key and that Shift is synthesized via the `sym` layer. That's wrong -- the T-Deck Plus keyboard has **two physical Shift keys**. While implementing this feature, please update `CLAUDE.md` to reflect the actual layout so future shortcut design isn't routed around a key that does exist.
- The lock should likely live as a top-level filter in `lua/ezui/init.lua` (or wherever the input dispatch loop sits) so it intercepts both `key/down` and touch events before any screen sees them.
- Touch swallowing needs to also prevent the touch driver in `lua/ezui/touch_input.lua` from forwarding events while locked.
- Hint overlay should use the existing transient/overlay primitives (`lua/ezui/transient.lua` or similar) so it draws on top of every screen including modals.
- Consider posting a `bus` event when lock state changes, so any service that wants to react (e.g. dim the backlight further, suppress sounds) can subscribe.

## Acceptance

- [ ] `Shift+Alt+L` from any screen enters lock mode and shows the hint.
- [ ] In lock mode, no keypress (other than the unlock chord) and no touch reaches any screen / widget.
- [ ] `Shift+Alt+U` exits lock mode and removes the hint.
- [ ] Mesh, GPS, and other background services continue normally while locked.
- [ ] Boot always starts unlocked.
- [ ] `CLAUDE.md`'s keyboard-layout section is corrected to acknowledge the two physical Shift keys.

xtr-bas:
Closes #77.

## Summary
- New `services/input_lock` module holds the lock state in memory, posts `input_lock/changed` on the bus, and resets to unlocked on every boot (matches the issue's "don't brick yourself with a stuck lock" requirement).
- `screen.handle_input` recognises **Shift+Alt+L** / **Shift+Alt+U** before everything else and, while locked, swallows every other keypress. The unlock chord still wakes the screensaver, so a single chord both wakes and unlocks.
- `touch_input.on_down` / `on_move` / `on_up` early-return when locked — taps, drag-to-scroll, and mouse-mode cursor moves are all dead in lock mode.
- A bottom-pinned banner ("Locked -- Shift+Alt+U to unlock") is drawn on top of every screen, including screensaver and mouse cursor, so the user always knows how to recover.
- Background services (mesh, GPS, notifications) keep running -- this is purely an input-layer gate.

## CLAUDE.md correction
- Removed the "no dedicated shift key, shift is synthesised via sym layer" claim. The T-Deck Plus matrix has two physical shift keys (`SHIFT1` / `SHIFT2` in `src/hardware/keyboard_matrix.h`), and either sets `key.shift`. Adjusted the Alt+Shift section to note that such chords are now appropriate for system-level global shortcuts.

## Test plan
- [x] `pio run` builds clean.
- [x] On-device: boot, confirm `input_lock.is_locked() == false`.
- [x] `-k l --shift --alt` sets `is_locked() == true`, banner appears at bottom.
- [x] While locked, `-k right`, `-k a`, `-k enter` are all swallowed (Desktop stays on Messages icon, no icon launch).
- [x] `-k u --shift --alt` clears the lock; banner disappears; subsequent navigation works again.
- [x] `input_lock/changed` bus event fires once per real state transition (deduped on no-op `set(true)`).
- [ ] On-device: confirm physical Shift+Alt+L and Shift+Alt+U chords work with both physical shift keys.
- [ ] On-device: confirm touch is gated (tap an icon while locked, nothing should happen).


xtr-bas:
Addressed in c53240b.

`locked_swallow` was `local`, so only the bridge's own `on_down`/`on_move`/`on_up` consulted it. Following the existing `is_wake_event` pattern:

- Exported `touch_input.is_locked()` mirroring `is_wake_event()`.
- Added the guard to every direct `touch/*` subscriber called out in the review and the rest of the codebase: `lua/screens/menu.lua`, `lua/screens/apps/paint.lua`, `lua/screens/apps/editor.lua`, `lua/screens/chat/contacts.lua`, `lua/screens/chat/messages.lua`, `lua/screens/tools/logs.lua`, `lua/screens/tools/touch_test.lua`, and `lua/screens/games/minesweeper.lua`'s `touch/move`.
- `touch/tap` and `touch/long_press` subscribers (minesweeper, solitaire) stay as-is -- those are synthesised inside the bridge's `on_up`, which already early-returns while locked, so they're covered transitively.

Extended the "Touch input and the screensaver wake gate" section in `CLAUDE.md` to document `is_locked()` alongside `is_wake_event()`, and updated the canonical subscriber snippet to call both predicates.

Verified on device:
- `touch_input.is_locked()` returns `true` when `input_lock.set(true)` is called, `false` after `set(false)`.
- Pre-fix bypass screens (Desktop tab strip, Paint canvas, etc.) now consult the predicate at the top of every subscriber.

xtr-bas:
Round-2 review addressed in cee3914.

**Code bugs**
- `minesweeper.lua:on_touch_move` and all three `touch_test.lua` handlers now check `is_locked() or is_wake_event()`, matching the canonical pattern in the updated CLAUDE.md.
- `logs.lua` `touch/up` now always runs `drag = nil` regardless of lock state, so a lock-mid-drag can't leave a stale anchor that produces a spurious scroll on the next post-unlock `touch/move`. Audited the other touch handlers I added: `menu.lua`'s `touch/up` already clears `pending` inside its locked branch; the rest don't carry between-event state.

**CLAUDE.md drift**
- Added `input_lock` to the `lua/services/` line in Project Structure.
- Added `input_lock` to the "loaded on demand" sentence alongside `map_archive`, `prefs_registry`, `signal_test`.
- New "Input lock service" subsection mirroring the Notifications service block: public API (`is_locked`/`set`/`toggle`), bus topic `input_lock/changed` with `{ locked = bool }`, the chord, the in-memory-only / boot-resets design note, and the cross-reference to the touch-gate opt-in contract.

**User-facing docs**
- New `lua/docs/manual/input-lock.md` (embedded into firmware, visible on-device via Help) covering what the feature does, the chord, what does and doesn't keep running while locked, and the "power cycle is the unconditional escape hatch" reset note.
- Linked from `lua/docs/manual/index.md` and chord added to the Global table in `lua/docs/manual/key-bindings.md`.

Verified on device: `ez.docs.list()` now includes `@manual/input-lock.md`, and the lock/unlock chord still round-trips cleanly.

Issue #122, opened 2026-05-13
feat(chat): multi-party group chat with rotating key


bvdaakster:
## Problem

The radio stack records RSSI and SNR for every received packet (`src/mesh/meshcore.cpp:158`, surfaced to Lua at `lua/services/custom_packets.lua:297`), but users have no way to visualise link quality on the map. They cannot answer "will my packet actually reach that repeater from here?"

> Note: this issue depends on repeaters being rendered on the map in the first place, which they currently are NOT. See the separate "show known peers / repeaters as pins on the map" issue. Until that lands, this overlay would render rings on an empty map. Both can be designed together.

## Proposal

A toggleable overlay on the map that draws a coarse coverage ring around each known peer (especially repeaters) with a recent location, sized by observed link quality.

### Data

- For each known peer with a location: keep a small ring buffer of the last N (start with 16) RSSI samples taken when the peer's signal was received.
- Smooth (median, not mean -- robust to outliers).
- Map the smoothed value to a coarse "good / marginal / poor" bucket. The actual distance the link spans is unknowable without two-way ranging, so the radius represents the user's empirical confidence -- it's the distance from the current GPS fix to the peer, scaled by signal quality.

### Rendering

- The map widget's existing `overlay_fn` hook (`lua/ezui/widgets/map_view.lua:234`) draws three concentric rings per peer:
  - Inner solid ring: distance peer-to-here, when we've heard them strongly.
  - Middle dashed ring: estimated marginal radius.
  - Outer dotted ring: estimated poor-quality radius.
- Toggled on/off from the map's M-key menu.
- Color uses the accent palette so it's readable in both dark and light themes (CLAUDE.md theming rule -- use `theme.color()` not raw RGB565).

### Honest about limits

This is a heuristic -- it cannot predict where the user can reach, only describe where they have reached. Label the overlay "Observed coverage" not "Predicted coverage". Hide the rings for any peer with fewer than 4 samples in the last hour.

## Out of scope

- Path-loss modelling (terrain / Fresnel zones). Out of scope on a T-Deck -- send people to a host-side tool if they want that.
- Coverage history visualisation across days.

## Files

- `lua/services/link_quality.lua` (new) -- ring buffers per peer.
- `lua/ezui/widgets/map_view.lua` -- nothing changes; overlay is a caller concern.
- `lua/screens/tools/map.lua` -- toggle + overlay function.

PR #126, opened 2026-05-13
feat(map): render known peers as pins on the map

### LINKED ISSUE #125: feat(map): show known peers and repeaters as pins on the map
## Problem

The map shows the user's own GPS fix (`lua/screens/tools/map.lua` via `make_gps_overlay()`), but it shows nothing else. Known peers -- repeaters, room servers, contacts whose recent ADVERTs included a location -- exist in memory and on disk, but never appear on the map. As a result the user has no spatial sense of who is around them, where the closest repeater is, or which contacts share their general area.

This is also a hard dependency for the range-estimator overlay (#122) -- coverage rings can't render around peers that aren't drawn.

## Proposal

Render every known peer with a recent location as a pin on the map.

### Sources

ADVERT packets carry an optional lat/lon (CLAUDE.md "App Data Structure", flags bit 4 = 0x10 "Has location"). Today these coords are parsed but only used for contact detail display. Three sources to surface:

- **Repeaters** (advert flags bits 0-1 = 0x02). Different glyph -- antenna icon.
- **Room servers** (flags 0x03). Building / hub icon.
- **Contacts** (flags 0x01) -- only when the user has actually added them as a contact, to keep `#Public` strangers from cluttering the map. Configurable: "Show all chat nodes" toggle for users who want fuller situational awareness.

### Rendering

- Use the existing `overlay_fn` hook (`lua/ezui/widgets/map_view.lua:234`) so it composes with any other overlays (GPS dot, range rings, GPS shares).
- Pin label = peer name truncated to ~10 chars; ASCII-sanitized per the fonts rule (peer-originated data).
- Stale data: peers whose last ADVERT was > 24 h ago render dimmer; > 7 days hide entirely (with a "Show stale peers" toggle in the M-key menu).
- Tap a pin -> small popup with name, role, distance from current fix, last-seen relative time, and an action menu ("Send DM", "Show on map" centered, "Remove from map" -- which just hides it locally).

### Storage / state

No new on-disk format -- the contacts and known-nodes stores already track lat/lon and last-seen. Add a `get_nodes_with_location()` helper on whichever service owns the discovered-nodes list.

### Settings

Settings -> Map -> "Show on map":
- "My location" (today's behaviour, keep togglable).
- "Repeaters and room servers" (default on).
- "My contacts" (default on).
- "All chat nodes" (default off -- privacy-aware default).

## Out of scope

- Trail / movement history for peers (separate, harder; needs persistent history).
- Crowdsourced repeater submission (the on-device list is what we know from our own ADVERTs).

## Files

- `lua/screens/tools/map.lua` -- add the peer overlay layer alongside GPS overlay.
- `lua/services/contacts.lua` / wherever discovered nodes live -- `get_nodes_with_location()` helper.
- `lua/screens/settings/map.lua` (new or extend) -- visibility toggles.

xtr-bas:
## Summary

Render every mesh peer with a known ADVERT location as a pin on the offline map. Repeaters draw as triangles, room servers as squares, and chat-node contacts as circles, each with a 10-char ASCII-sanitized name label. Peers older than 24h render dimmer; peers older than 7d are hidden.

A new `Settings -> Map` sub-page toggles each peer class (repeaters/room servers, my contacts, all chat nodes, show-stale). Defaults favour situational awareness for infrastructure and contacts while leaving unknown chat nodes off for privacy.

Unblocks the range-estimator overlay (#122), which needs peer pins to anchor coverage rings.

## Test plan

- [x] `pio run` clean (RAM 26.2%, flash 58.4%)
- [x] Flashed to T-Deck Plus; opened the Amsterdam archive with the user's CLI repeater (`NL-IJM-1`) in view, confirmed the triangle pin + name label render in the correct geographic position
- [x] Toggled `map_peer_inf` off via `ez.storage.set_pref`; pin disappeared on next frame. Re-enabled; pin reappeared
- [x] No errors in `--logs` after exercising the Map screen and the new Settings -> Map screen

## Out of scope

The issue mentions a tap-pin popup (DM / show-on-map / hide locally) and an M-key "Show stale peers" action. Both are deferred to follow-ups so this PR can land the core pin-rendering surface that #122 depends on. The stale-peers toggle is already available in `Settings -> Map` for now.

Closes #125


xtr-bas:
Doc page added in 90d2249.

Created at `lua/docs/manual/settings/map.md` (instead of the suggested `docs/manual/settings/map/index.md`) -- per CLAUDE.md's source-of-truth split, `lua/docs/manual/` is the hand-authored tree that gets embedded into firmware and rendered on-device by the Help app; `docs/manual/` is the auto-generated tree (`Do not hand-edit; deployed to GitHub Pages`).

The embedder under `scripts/embed_lua_scripts.py` `rglob`s `lua/docs/`, so the new file gets virtual path `@manual/settings/map.md` and `ez.docs.list()` surfaces it automatically on the Help screen -- no registry update needed.

Content covers all four toggles, defaults, the 24h/7d staleness thresholds, and the privacy reasoning for keeping "All chat nodes" off by default.

PR #142, opened 2026-05-15
chore: audit Lua key handlers for keys unreachable on T-Deck Plus

### LINKED ISSUE #53: chore: audit Lua key handlers for keys not on the T-Deck Plus physical keyboard
## Background

The T-Deck Plus keyboard has no number row, no Ctrl, no Esc, no Tab, no function keys. CLAUDE.md's "Keyboard layout" section spells this out:

> Numbers and most punctuation are entered via \`alt+letter\` combinations, NOT via a number row.
> Never use \`key.ctrl\` for an on-device shortcut. The remote tool can send it, but a user in the field cannot.
> Never rely on \`ESCAPE\` as the only exit path.

PR #50's video poker bound holds to bare digits 1-5 -- typable via the remote tool but unreachable as plain keypresses on-device. Caught by claude[bot] review on the PR (now fixed in 1113368). Likely not the only place this slipped in.

## Asks

Audit every \`handle_key\` (and any direct \`key.character\` / \`key.special\` checks elsewhere) under \`lua/\` for:

1. **Bare digit characters** (\`key.character == "0".."9"\`) used as primary on-device input. These only arrive via alt-chord on the T-Deck, so the user-visible hint should say so OR the binding should move to a letter.
2. **\`key.ctrl\` checks** with no \`key.alt\` alternative. \`Ctrl\` doesn't exist on the device.
3. **\`key.special == "ESCAPE"\` as the sole exit path** with no \`BACKSPACE\` synonym. There's no Esc key on the device; back-arrow sends BACKSPACE.
4. **Bare punctuation characters** (\`-\`, \`=\`, \`/\`, etc.) used as shortcuts -- same alt-chord problem as digits.
5. **Hint text that mentions \`Ctrl\`, \`Esc\`, or bare digits** even when the binding itself is fine -- the on-screen instruction is wrong.

Output: a punch list per file with line numbers and the suggested fix. Not implementation -- the survey alone is what's useful here, then we triage.

## Notes

- Per CLAUDE.md the only modifier reachable from the keyboard is \`alt\`. \`shift\` is generated by the \`sym\` layer.
- Trackball acts like arrow keys, so anything bound to UP/DOWN/LEFT/RIGHT is fine.
- Some screens may legitimately want Ctrl/Esc support for the remote tool *in addition to* the on-device binding -- those aren't bugs, but the on-device path needs to exist.

xtr-bas:
## Summary
Surveyed every \`handle_key\` and ad-hoc \`key.character\` / \`key.special\` check under \`lua/\` for bindings or hint text that depend on Ctrl / Esc / Tab / bare digits -- keys the T-Deck Plus physical keyboard cannot produce on its own. Output is a triage punch list in \`docs/audits/key-handler-audit.md\`; no code changes, per the issue (the survey is the deliverable).

Headlines:
- **0** \`key.ctrl\` uses in Lua (already clean).
- **1** user-visible hint mentions "Esc" (\`solitaire.lua:607\`).
- **2** \`TAB\` bindings; one (\`file_manager.lua:522\`) has no on-device alternative.
- **2** help screens binding bare digits without alt-chord guidance (\`wasteland\`, \`sudoku\`).
- Source-comment cleanup list deferred to future passes.

Closes #53

## Test plan
- [ ] **Needs hardware QA** — not strictly required (docs-only PR) but worth a sanity build: \`pio run -t upload\` from this worktree and confirm boot.
- [ ] Read \`docs/audits/key-handler-audit.md\` and decide which findings to spin into follow-up issues / PRs.
- [x] Build is clean (\`pio run\` succeeds, verified locally).

### DIFF
--- a/docs/audits/key-handler-audit.md
+++ b/docs/audits/key-handler-audit.md
@@ -1,0 +1,179 @@
+# T-Deck Plus Lua key-handler audit (#53)
+
+Survey of `handle_key` and ad-hoc `key.character` / `key.special` checks
+under `lua/` looking for bindings or hint text that depend on keys not
+reachable from the T-Deck Plus physical keyboard. Source of the rules:
+the "Keyboard layout (T-Deck Plus)" section of CLAUDE.md.
+
+The on-device keyboard has:
+
+- letters A-Z, plus the dollar key
+- `space`, `Enter`, `Backspace`
+- `alt`, `sym`
+- two black side keys (mic / speaker)
+- trackball (UP/DOWN/LEFT/RIGHT/ENTER from click)
+
+No Ctrl, no Esc, no Tab, no function keys, no number row. Numbers and
+most punctuation only arrive via `alt+letter` chords.
+
+Scope of this pass: bindings that block a user-reachable action on the
+T-Deck. Remote-tool-only synonyms (e.g. ESCAPE alongside BACKSPACE,
+Ctrl in dev-only paths) are not findings.
+
+## Summary of findings
+
+- **0 instances of `key.ctrl`** in Lua. No work needed there.
+- **6 `ESCAPE`-only files** -- 5 of them also accept `q` so the user
+  has a reachable exit. 1 (`map_loader.lua`) has a usable `q`.
+- **1 user-visible hint mentions "Esc"** (`solitaire.lua:607`).
+- **2 user-visible help screens binding bare digits** without
+  documenting the alt-chord (`wasteland.lua:2343`, `sudoku.lua` --
+  digits are part of the gameplay loop, in-game help missing).
+- **2 `TAB` bindings** without a clearly documented on-device
+  alternative (`desktop.lua:527`, `file_manager.lua:522`). Desktop
+  also accepts `m`; file_manager has none.
+- Numerous **comment-only** mentions of ESC/Ctrl in source comments
+  -- not user-visible, but worth scrubbing during the next pass over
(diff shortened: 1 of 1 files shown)


xtr-bas:
Addressed all five inline comments in \`cc286bd\`:

- **\`docs/audits/key-handler-audit.md:16-17\`** -- corrected the blanket "alt+letter only" intro to call out \`0\` as the dedicated Row 4 key.
- **\`docs/audits/key-handler-audit.md:26-27\`** -- rewrote the self-contradictory summary line to "all 6 also accept \`q\`".
- **\`docs/audits/key-handler-audit.md:59-61\`** -- sudoku hint suggestion now reads \`alt+1..9 fill, 0/Back clear\`, with a note that BACKSPACE already clears via the handler at \`sudoku.lua:274\`.
- **\`docs/audits/key-handler-audit.md:111-122\`** -- dropped the misleading option 2 (\`q:cancel\`). \`q\` returns \`"pop"\` unconditionally at \`solitaire.lua:685\` -- it quits, never cancels. Only the BACKSPACE-handler fix remains.
- **\`docs/audits/key-handler-audit.md:149\`** -- image_viewer header-comment suggestion changed to \`"... Back quits"\`; \`q\` is not wired in the handler (\`image_viewer.lua:167-194\` only exits on BACKSPACE/ESCAPE).

\`83c482f\` merges latest \`origin/test\` into the branch; \`pio run\` clean.

Issue #127, opened 2026-05-13
feat(mesh): broadcast a user-picked home location in our ADVERTs


bvdaakster:
## Problem

The MeshCore ADVERT protocol natively supports a location field (CLAUDE.md "App Data Structure": flags bit 0x10 = "Has location" + `[lat_e6:4 LE][lon_e6:4 LE]` in app_data). Our parser already consumes incoming GPS-tagged adverts -- `src/mesh/meshcore.cpp:323-331` stamps the node's coords into our known-nodes store, which is how we know where some repeaters are.

But our outgoing path doesn't reciprocate. `MeshCore::sendAnnounce()` at `src/mesh/meshcore.cpp:600` hardcodes `flags = 0x81` (client + has_name) and never sets the location bit. Other nodes see our name but not our position, even on the map.

This is a feature gap that would meaningfully enrich #125 (peer pins on map) and the #122 range estimator -- but doing it naively (broadcasting live GPS) is a real privacy hazard.

## Scope: home-base mode only

After working through the threat model (trajectory averaging attacks, snap-to-grid leakage, rural cell intersection, long-term aggregation, RSSI triangulation side channel), we concluded that any form of live or fuzzed live broadcast is privacy theatre against a motivated adversary. The only honest framing is:

**The user picks one fixed point that gets broadcast unchanged.**

This is a deliberate lie they author themselves. No runtime computation, no per-advert randomization, no fuzz radius (which gives wildly different real-world privacy depending on user density and falls to averaging attacks anyway). One coordinate, picked once, broadcast every time.

## UI

No dedicated picker screen, no Settings entry, no link from Privacy. The existing Map screen (`lua/screens/tools/map.lua`) already has everything needed: it tracks `center_lat` / `center_lon`, the user pans / zooms with the trackball + keys, and the centre of the viewport is implicitly a crosshair.

Add a single Alt+M action on the Map: **"Set as broadcast home"**. The action:

1. Reads the current `center_lat` / `center_lon` from the map state.
2. Writes `adv_home_lat` / `adv_home_lon` to NVS (decimal-e6 strings).
3. Posts a one-shot confirmation notification ("Broadcast home set. Anyone receiving your adverts will see this coordinate -- treat it as public.").

[...]

PR #158, opened 2026-05-15
feat(sharing): gps/v1 share-card for location over DM or channel

### LINKED ISSUE #106: feat(sharing): share GPS coordinates via DM or channel
## Problem

The plumbing to share a GPS fix over the mesh already exists -- `custom_packets.send_gps()` packs a 12-byte record (lat/lon e6, alt, sats, flags) and `custom/gps_fix` fires on the receiving side (`lua/services/custom_packets.lua:384` / `:460`) -- but **nothing in the UI invokes it**, and there is no chat-bubble representation. Users can see their own location on the map, but they cannot say "I'm here" to a contact or to a channel.

We also have no `ezmesh://` share-card URI type for a coordinate, so positions can't be passed through the share/parse round-trip the way contacts, channel invites, time, and sigt pings already are (`lua/services/sharing.lua:37-41`).

## Proposal

Add a `gps/v1` share-card type plus the UI to send and receive it.

### URI design (mirrors existing types)

- **DM (encrypted):** `https://ezme.sh/#gps/v1?t=<base64url(nonce8 || AES128_ECB(secret, blob))>` -- same X25519-scoped pattern as `join/v1` so coords aren't leaked in the rendezvous URL. The blob carries `[lat_e6:4 LE][lon_e6:4 LE][alt_m:2 LE][flags:1][label:N]`, capped to fit the 120-char DM payload budget.
- **Channel (plaintext):** `https://ezme.sh/#gps/v1?lat=<e6>&lon=<e6>&n=<urlencoded label>` -- channel members already see each other's traffic in cleartext, so encrypting is pointless overhead.

A short numeric `pos/v1` variant for "I'm here right now" ephemeral pings (no label, not surfaced as a pin) could share the same parser.

### UI

- **Map screen action menu (M-key):** "Share location -> DM..." / "Share location -> Channel..." -- opens contact / channel picker, then drops the URI into the conversation via the same `sharing.send_*` flow contacts already use.
- **Chat compose:** "Attach location" entry in the compose action menu, same destinations.
- **Inbound bubble:** rendered like any other share card -- title is the label (or "Shared location"), subtitle is `lat, lon` to 4dp. Per CLAUDE.md, **all** bubble actions go behind the context menu: "Show on map" (push map screen centered on the point + drop a pin via the existing `overlay_fn` hook in `lua/ezui/widgets/map_view.lua:232`), "Copy coordinates", "Navigate to" if/when we have a bearing helper.
- **Pin persistence:** received coords from `custom/gps_fix` and `gps/v1` cards should optionally drop a labelled pin on the map. Start with an in-RAM ring buffer keyed by sender pubkey hash; persistence can come later with the waypoints feature.

### Safety / privacy

- Respect the `gps_enabled` pref -- never auto-share, every send is an explicit user action.
- Sanitize the label to printable ASCII (CLAUDE.md "On-device font character set"); it can come from a peer.
- Add a notification source tag (`notify_gpsshare`) so the toast is mutable.

## Out of scope

- Continuous live-location streaming (separate feature; see the route-recording issue).
- Waypoint library / favourites.

## Files likely touched

- `lua/services/sharing.lua` -- new `build_gps_card` / `parse_gps_card` + add `gps/v1` to the dispatch table.
- `lua/services/custom_packets.lua` -- already done; just make sure the bus event seeds the on-map pin store.
- `lua/screens/tools/map.lua` -- action menu entries + pin overlay layer.
- `lua/screens/chat/*` -- compose action menu + inbound bubble action menu.
- `lua/docs/manual/sharing/` -- document the new URI type.

xtr-bas:
## Summary

Adds the issue #106 \`gps/v1\` share-card URI in two flavors:

- **Channel** (\`?lat=&lon=&n=\`): plaintext, since channel members already see every message in the clear -- wrapping is pointless overhead.
- **DM** (\`?t=<base64url(nonce || aes128_ecb(x25519_secret, blob))>\`): same X25519-scoped envelope as \`#join/v1\`. Blob = \`[lat_e6:4 LE i32][lon_e6:4 LE i32][label_len:1][label:N]\` padded to one AES block. Cap on label is 32 bytes, picked to fit the resulting URL inside DM's 120-char text budget.

\`services.sharing\` exposes \`encode_gps_channel\`, \`encode_gps_dm\`, \`decode_gps_dm\`, and \`parse()\` recognises the new verb. A small \`screens/chat/gps_share\` module mirrors \`time_share\`'s shape -- both DM and channel chats wire it for bubble cards (LOCATION kind label, lat/lon hint) and a context menu with **Show on map** (pushes the Map screen centered on the point at zoom 14, clamped to the archive's zoom range) and **Copy coordinates** (confirmation dialog with the precise lat/lon, since there's no system clipboard).

## Send paths

- **Chat Alt+M -> Share my location.** Uses the current GPS fix if available, else the user's broadcast-home pref as a fallback (the user already accepted that point is public-grade in #127). DM conversations get the encrypted variant; channel conversations get the plaintext one.
- **Map Alt+M -> Share this point -> DM... / Channel...**. Pans the crosshair to an arbitrary point and picks a destination via a list dialog (same shape as the contact-share picker in DM). Reuses the existing share-send pipeline.

## Receive paths

The existing binary GPS\0 custom-packet path (\`services/custom_packets\`) already fires \`custom/gps_fix\`; this PR adds a \`boot.lua\` subscriber so inbound coords surface as a notification toast under the new \`gps\` source. Mute via the \`notify_gps\` NVS pref (\"0\" = silent), per the same convention as the existing DM / file / battery sources.

URL-shaped GPS shares (the new path) render as a LOCATION chat bubble; tapping it opens the standard context menu. Self-sent and undecryptable variants show their own disabled card so the sender / failure state is visible.

## Out of scope (per the issue)

- **On-map pin persistence for received shares.** The issue marks this as a follow-up tied to the waypoints feature; the \"Show on map\" action centers the map but doesn't drop a labelled pin. Worth a follow-up issue if you want the pin layer.
- **Long-press direct trigger on chat bubble.** CLAUDE.md is explicit that every bubble action must be behind the context menu; tap-to-trigger would violate that.
- **Settings panel for \`notify_gps\` mute.** The pref-source namespace exists; a settings entry across all sources is a separate sweep.

## Test plan

- [x] \`pio run\` builds clean on this branch (RAM 26.2%, Flash 59.1%).
- [ ] **Needs hardware QA** -- \`pio run -t upload\` from this worktree, then exercise:
  - DM conversation, Alt+M -> Share my location with a contact you share a secret with. Receiver sees a LOCATION bubble, opens the context menu, picks Show on map -> map opens centered on the point.
  - DM conversation receiving a self-sent location share. Bubble renders as \"(location sent)\" -- self path can't decrypt its own DM token.
  - Channel conversation, Alt+M -> Share my location. Plaintext URL appears in the bubble; every channel member sees the LOCATION card and can Show on map.
  - Map screen, Alt+M -> Share this point -> DM... -> pick contact. Recipient gets a LOCATION bubble.
  - Map screen, Alt+M -> Share this point -> Channel... -> pick channel. Channel sees the LOCATION bubble.
  - Confirm \"Share my location\" falls back to broadcast home when there's no GPS fix, and surfaces a system bubble / toast when neither is available.
  - Send a binary GPS\0 packet from a peer device (\`custom_packets.send_gps\`) and confirm a toast appears under the \`gps\` source. Set \`notify_gps=0\` and confirm the toast is suppressed.
  - Compose menus on existing DM / channel screens still send contact-share, channel-invite, and time-share URLs without regression.

Closes #106


xtr-bas:
Addressed in \`24240f2\`:

- **\`lua/screens/chat/gps_share.lua:136\` (unsanitized sender_name)** -- wrap \`msg.sender_name\` in \`ascii_safe\` to match the sibling \`label\`/\`r.label\` guards. Peer-originated names with non-ASCII bytes now render as \`?\` substitutions instead of \`[]\` boxes.
- **\`lua/services/sharing.lua:213\` (asymmetric rounding)** -- replaced the \`(v >= 0 and 0.5 or -0.5)\` branch with a uniform \`+ 0.5\`. \`to_e6(-90.0)\` now yields \`-90000000\` (was \`-90000001\`), so the encoder no longer rejects exact negative-boundary values. Updated the intent comment to spell out why a symmetric offset is wrong here.

\`8c485f9\` is just \`origin/test\` merged in (CLEAN auto-merge); \`pio run\` clean.

PR #159, opened 2026-05-15
feat(map): record GPS route and replay it on the map

### LINKED ISSUE #108: feat(map): record GPS route and replay it on the map
## Problem

GPS works (`lua/services/gps.lua`) and the map widget already has the overlay hook needed to draw arbitrary geometry on top of tiles (`lua/ezui/widgets/map_view.lua:232` -- `overlay_fn(d, x, y, w, h, project)` where `project(lat, lon) -> px,py`). What's missing is the actual track recorder: a background sampler that turns the user's movement into a polyline they can review, save, and (eventually) share.

### Recorder (`lua/services/gps_track.lua`, new)

- `start({ label?, min_interval_s?, min_distance_m? })` -- begins a session.
- `stop()` -- finalises and writes to `/sd/tracks/<unix_ts>-<label>.eztrack`.
- `current()` -- returns the in-progress polyline for the live overlay.
- `list()` / `load(path)` -- enumerate and replay saved tracks.
- Sampling rule: take a new point when `delta_t >= min_interval_s` **and** `haversine(prev, now) >= min_distance_m` (defaults 5 s / 5 m, configurable in Settings). This is the standard "Douglas-Peuckerable input" trick -- keeps the file tiny without losing shape.
- Auto-pause when GPS loses fix; resume on re-acquire (gap is preserved as a polyline break, not interpolated).
- Survives reboot: append-only file format with a 4-byte length-prefixed header, then `[ts_delta:2][lat_e6:4][lon_e6:4][alt:2][hdop:1]` records (~13 B/point). A crashed session is recoverable from the file on next boot.

### Storage format (`.eztrack` v1)

```
magic "EZTRK1" (6)
flags (1)             # bit0 = closed/finalised
start_unix (4 LE)
label_len (1) + label (N, ASCII)
[records...]
```

Keep it append-only so the foreground recorder never has to rewrite. A "compact" pass on stop can run Douglas-Peucker to shrink for sharing.

### UI

- **Map screen action menu:** "Start recording route" / "Stop recording" / "Open saved track..."
- **Live overlay:** the in-progress polyline draws on top of tiles via the existing `overlay_fn`, in the theme accent colour, with a fading head so the user can tell direction at a glance.
- **Track viewer screen:** push from "Open saved track..." -- lists `.eztrack` files, opens one onto the map with auto-fit bounds, M-key actions: "Rename", "Delete", "Share..." (future), "Stats" (distance / duration / avg speed / elevation gain when available).
- **Recording indicator:** small red dot in the title bar of any screen while a session is active, like the desktop currently shows the GPS state. Tap to jump to the map.

### Settings

- Settings -> GPS -> "Track recording": min interval, min distance, autostart-on-boot toggle (off by default -- privacy).

### Safety / privacy

- Respect `gps_enabled`. If the user disables GPS mid-session, stop and finalise cleanly.
- No automatic mesh transmission. Sharing a track is a separate (future) feature; recording is local-only by default.
- Files live on SD so they can be pulled off the device via USB without round-tripping mesh.

## Out of scope

- Streaming a live route to peers in real time (separate feature; ties into the GPS-share issue).
- GPX export -- nice to have, but a host-side tool can convert `.eztrack` later; on-device space is the priority.
- Multi-track overlay diff / merge UI.

## Files likely touched

- `lua/services/gps_track.lua` (new)
- `lua/screens/tools/map.lua` -- menu items, live overlay, recording indicator.
- `lua/screens/tools/track_viewer.lua` (new)
- `lua/screens/settings/gps.lua` -- recorder prefs.
- `lua/boot.lua` -- start the service so a crashed session can be reaped.
- `lua/docs/manual/map/` -- document the recorder.

xtr-bas:
## Summary

Adds the issue #108 track recorder + viewer. The recorder is a long-lived sampler coroutine gated by min-interval **and** min-distance thresholds (defaults 5 s / 5 m, configurable in Settings). Every captured fix is appended to a `.eztrack` file under \`/sd/tracks/\`; the in-progress polyline draws live on the map.

## What's in the PR

- **\`services/gps_track\`** (new). API: \`start({ label?, min_interval_s?, min_distance_m? })\`, \`stop()\`, \`current()\`, \`list()\`, \`load(path)\`, \`delete(path)\`, \`live_points()\`, \`reap_unfinalised()\`, \`get_prefs() / set_prefs()\`. Sampler runs once at module load, polls at 1 Hz, and respects the active flag without re-spawning.
- **\`.eztrack\` v1 format**:
  ```
  \"EZTRK1\" (6)
  flags (1)         # bit0 = closed/finalised
  reserved (1)
  start_unix (4 LE)
  label_len (1) + label (N, ASCII)
  records...        # 13 B each: ts_delta:2 LE u16, lat_e6:4 LE i32,
                    # lon_e6:4 LE i32, alt_m:2 LE i16, hdop*10:1 u8
  ```
- **Map screen** (\`screens/tools/map.lua\`). Alt+M gains \"Start recording route\" / \"Stop recording route\" / \"Open saved track...\". The \`overlay_fn\` composes a polyline layer (saved track from the viewer, or live points from the recorder) under the GPS dot. Status strip adds a \`REC\` badge while a session is active. \`Map:update()\` invalidates every ~1 s during recording so the live tail extends without a full state rebuild.
- **Track viewer** (\`screens/tools/track_viewer.lua\`, new). Lists tracks newest-first with size + relative-time-ago + closed status. Alt+M actions: Open on map (auto-fits viewport via lat/lon span), Stats (point count + km + duration), Delete (confirmation dialog).
- **Settings** (\`screens/settings/gps_settings.lua\`). New \"Track recording\" section with min-interval / min-distance dropdowns. Same defaults as the recorder.
- **Reaper** (\`lua/boot.lua\`). A spawn at boot calls \`gps_track.reap_unfinalised()\` so any session that lost power mid-record has its closed bit set on next boot. The viewer otherwise would forever show \"(in progress)\" for those files.

## Privacy / safety

- \`start()\` refuses when GPS is disabled in Settings or when the system clock isn't set (a track with zero \`start_unix\` is useless).
- Nothing is transmitted automatically. Sharing a track would route through file_transfer, which is a separate flow.
- Tracks live on SD so they can be pulled off via USB without round-tripping the mesh.
- The Map screen's REC badge is the running indicator. There's no global indicator outside the map screen yet -- a follow-up could add one to the title-bar or status area.

## Deliberately deferred

- **Global recording indicator** outside the Map screen. Worth a follow-up issue tied to a shared \"status badges\" mechanism rather than threading state through every screen.
- **Autostart-on-boot pref** -- per the issue, off by default for privacy; a pref with no UI hook plus no Settings entry is mostly a foot-gun, so I left it out for now.
- **Douglas-Peucker compaction on stop** -- the live file already gets the size benefit of the (5 s / 5 m) threshold; a separate \"compact\" pass is easy to bolt on later.
- **GPX export** -- a host-side tool can convert \`.eztrack\` -> GPX more flexibly; on-device space is the priority.

## Test plan

- [x] \`pio run\` builds clean on this branch (RAM 26.2%, Flash 59.2%).
- [ ] **Needs hardware QA** -- \`pio run -t upload\` from this worktree, then:
  - Open Map, Alt+M -> \"Start recording route\". Status strip gains a \`REC\` badge; walking around makes the live polyline extend in the accent colour.
  - Alt+M -> \"Stop recording route\". \`REC\` badge clears; notification confirms point count. Pull SD, confirm a \`/sd/tracks/<unix>-track.eztrack\` exists with the closed bit set.
  - Alt+M -> \"Open saved track...\" lists the new file; pressing Enter opens the polyline on the map auto-fitted to its bounds.
  - Alt+M on a track row -> Stats shows point count / km / duration; Delete prompts and removes the file.
  - Settings -> GPS -> Track recording. Change min-interval and min-distance; start a new session and confirm the sampling rule visibly changes (e.g. 1 s / 0 m -> dense polyline, 30 s / 25 m -> sparse).
  - Walk into a building until the fix drops, then back out. Confirm the resulting polyline has a visible gap (no interpolation across the lost fix), per the issue's segment-break requirement.
  - Reboot mid-record. Boot, Alt+M -> \"Open saved track...\". The interrupted file shows up cleanly (no \"(in progress)\" suffix) because the reaper finalised it.
  - With GPS disabled in Settings, Start refuses and surfaces a system notification.

Closes #108


xtr-bas:
Addressed two of the three review comments in \`a5b1126\`:

- **\`lua/screens/tools/track_viewer.lua:258\` (Alt+M context menu)** -- removed the bare \`m\`/\`M\` block from \`Viewer:handle_key\` and added a \`Viewer:menu()\` returning the action items for the currently-focused track. The framework dispatches it on Alt+M, same path \`Map:menu()\` uses. Stats computation lifted into a module-level helper since it no longer captures the menu screen's \`self\`. Hint text updated from \`"M: actions"\` to \`"Alt+M: actions"\`.
- **\`lua/services/gps_track.lua:35\` (CLAUDE.md docs)** -- new \"GPS track recordings (.eztrack v1)\" subsection under Key Components, mirroring the Node Store entry: magic / flags / record struct / storage path / sampling prefs / 1 MB read_file ceiling on the finalise-on-disk path.

\`04fe5ad\` is just \`origin/test\` merged in (CLEAN auto-merge); \`pio run\` clean.

**Still open**: \`lua/services/gps_track.lua:400\` (large-track silent fail). The reviewer's two suggested paths -- add a Lua-side \`ez.storage.write_at\` binding so we can patch the flags byte without reading the whole file, OR cap \`start()\` so a session can't outgrow the read_file ceiling -- both involve decisions beyond a typical review iteration (new C++ surface vs a hard user-facing limit). Leaving that for a follow-up so this PR can land its current scope.

xtr-bas:
Follow-up: addressed the large-track silent-fail in \`ba54536\`.

- **\`src/lua/bindings/storage_bindings.cpp\`** -- new \`ez.storage.write_at(path, offset, data)\` opens the file in \`r+\` mode, seeks, and writes \`#data\` bytes over the existing content. Refuses to extend the file, refuses missing files. Mirrors the existing remote-control \`WRITE_AT\` command (0x0B).
- **\`lua/services/gps_track.lua\`** -- \`finalise_on_disk\` now patches the FLAG_CLOSED byte via \`write_at\` first. Read-rewrite kept as a fallback for the warm-OTA case where app Lua runs against pre-write_at C++.
- **\`CLAUDE.md\`** -- .eztrack section updated; the 22 h close-path ceiling is gone.

\`pio run\` clean.

xtr-bas:
Addressed in \`a04d0ad\`:

- **\`lua/services/gps_track.lua:408\` (non-atomic rewrite)** -- write_file fallback now writes to \`<path>.tmp\` and \`rename()\`s over the source. \`reap_unfinalised\` extended to sweep leftover \`.eztrack.tmp\` files on boot for the partial-write case.
- **\`lua/screens/tools/track_viewer.lua:132\` (extra dialog pop)** -- \`dialog.confirm\` pops its own screen before invoking \`on_ok\`, so the explicit \`function() screen_mod.pop() end\` ejected the user from the track list. Dropped from all three info dialogs (Empty track / Cannot read track / Stats).

\`pio run\` clean.

xtr-bas:
Addressed the inline review feedback in \`42ee891\` (preceded by a \`test\` merge in \`d3a00c8\`):

- **\`lua/services/gps_track.lua:213\`** (and the two sibling sites in \`read_header\` and \`M.load\`) -- replaced the \`A and B or C\` shorthand with an explicit branch so the \`read_file\` fallback only fires when \`read_bytes\` is genuinely absent. A nil return from either path now flows out via the existing "short read" / "read failed" error returns instead of crashing \`list()\` with a nil-deref.

\`pio run\` clean. Hardware-sensitive (SD file I/O paths); please flash and re-test list + load on real recordings (including one that exceeds the read_file 1 MB cap so the read_bytes path is the live one) before merging.

xtr-bas:
Addressed the three remaining inline asks in \`34cfad0\` (preceded by a clean \`test\` merge in \`da0b362\`); the \`.eztrack\` v1 on-disk format ask (gps_track.lua:35) was already covered by the existing CLAUDE.md section that the test merge carried in.

- **\`lua/screens/tools/track_viewer.lua:229\`** -- \`set_state({})\` merges no keys and leaves \`state.tracks\` stale; replaced with \`set_state({ tracks = gps_track.list() })\` so a deleted entry disappears without needing a screen pop/push.
- **\`lua/screens/settings/gps_settings.lua:294\`** -- added a \`### Track recording\` subsection to \`lua/docs/manual/settings.md\` under \`## GPS\` documenting the Min interval and Min distance dropdowns with the actual label/value pairs from the panel.
- **\`lua/boot.lua:546\`** -- inserted \`gps_track\` as item 13 in CLAUDE.md's Services init list (between \`gps\` and \`power\`, matching the wiring in \`boot.lua\`); renumbered \`power\` to 14.

\`pio run\` clean. UI-touching change (track viewer delete refresh); flash and re-test the delete flow before merging.

PR #160, opened 2026-05-15
feat(chat): emoji reactions over RAW_CUSTOM for DM messages

### LINKED ISSUE #112: feat(chat): message reactions and replies
## Problem

Chat is purely linear today -- no way to react to a specific message, no way to thread a reply. On a mesh where every packet costs airtime, "thumbs up" and "haha" are exactly the kind of low-information acknowledgement that should ride a tiny 2-3 byte payload, not a full TXT_MSG.

### Reactions

- New custom packet kind `reaction`: `[target_msg_hash:4][emoji_index:1]`. 5 bytes plus the MeshCore header -- well under `MAX_PACKET_PAYLOAD` (184).
- `target_msg_hash` = first 4 bytes of SHA-256 over `[sender_pubkey:32][timestamp:4][text]` of the message being reacted to. Computed locally; no extra metadata needed on the original.
- `emoji_index` = a small fixed palette (heart, thumbs-up, laugh, sad, fire, question, exclaim, check). The palette is firmware-defined so we don't pay per-message string cost.
- Reactions render as a small footer row on the target bubble, with counts when 2+ identical reactions arrive. Long-press on the footer lists who sent each.
- Bus event `chat/reaction` so screens can update incrementally.

### Replies

- Extend the local message record with an optional `reply_to_hash:4` field. On the wire, prepend `[flags:1 = HAS_REPLY][reply_to_hash:4]` inside the TXT_MSG inner plaintext when set. Existing receivers that don't know the flag will see the prefix as garbage in the text -- so we gate this behind a feature-advert flag, or pick a flag bit that older receivers will ignore (the current TXT_MSG inner plaintext flags byte already exists; see CLAUDE.md "Direct Messages" section).
- Reply bubbles render with a quoted-snippet header (truncated to one line, accent-colored bar on the left). Tapping the header scrolls to the original.

### UI

- Long-press a bubble (or M-key menu when focused) -> "React..." (opens the 8-emoji palette) and "Reply" (focuses compose with the quote attached).
- Reactions surface in the notifications service as a single source (`notify_reaction`) so they can be muted independently of DM toasts.

## Open questions

- Does the protocol bump need a version negotiation, or do we live with the "older receivers see no reaction" failure mode? Probably the latter -- reactions are intentionally lossy.
- Should reactions to channel messages flood, or be DM'd back to the sender? Flooding is simpler and matches the conversation model.

## Out of scope

- Custom emoji uploaded across the mesh.
- Edit / delete of a sent message (separate feature; harder because of distributed state).

## Files

- `lua/services/custom_packets.lua` -- new packet kind.
- `lua/services/direct_messages.lua` / `channels.lua` -- store + dispatch the reply-to field; surface reactions per message.
- `lua/screens/chat/*` -- bubble footer rendering, palette picker, reply quote rendering.
- `lua/services/notifications.lua` -- new source tag.

xtr-bas:
## Summary

Adds the reactions slice of issue #112: emoji ACKs for DM messages, riding the existing custom-packet (RAW_CUSTOM) layer instead of paying a full TXT_MSG round-trip. A reaction is 5 bytes plus envelope on the wire.

**Replies and channel-side are deferred to follow-up PRs** (see \"Out of scope\" below). Doing all three in one PR would explode the diff and bundle two independent wire-format changes; the reactions layer is the natural first slice because it's a clean new subtype with no compat risk.

## Wire format (\"RXN\\0\" subtype)

\`\`\`
data = [target_msg_hash:4][emoji_index:1]
target_msg_hash = sha256([sender_pubkey:32][timestamp:4 LE][text])[:4]
\`\`\`

Both peers can recompute \`target_msg_hash\` from data already in their copy of the message; no extra wire bits ride on the original TXT_MSG. RAW_CUSTOM is direct-range only (relays don't re-flood it), which is fine for 1-on-1 DMs.

## Emoji palette

Eight ASCII-only tokens, per CLAUDE.md's font character set rule:
\`+1\`, \`<3\`, \`:D\`, \`:(\`, \`^_^\`, \`!!\`, \`??\`, \`OK\`. A real heart-glyph would paint as \`[]\` on the default fonts; ASCII is honest and portable.

## Components

- \`services/reactions\` (new). In-RAM store keyed by (conversation_partner, target_msg_hash). Public API: \`compute_msg_hash\`, \`send\`, \`get_for\`, \`compress\` (counts + senders, descending), \`init\` (registers the custom-packet handler). Posts \`chat/reaction\` on the bus for inbound and self events.
- \`screens/chat/reactions_picker\` (new). List-row palette picker; pops back to the chat on selection so the optimistic local record + bubble rebuild show the reaction immediately.
- \`screens/chat/chat_common\`. \`chat_bubble\` measure + draw extended with a reactions footer row under the meta line. Bubble widens to fit a long row if needed.
- \`screens/chat/dm_conversation\`. \"React...\" entry in the bubble context menu; computes \`target_msg_hash\` per bubble; subscribes to \`chat/reaction\` for live updates.
- \`boot.lua\`. \`reactions.init()\` after \`custom.init()\`; \`chat/reaction\` notification under a new \`rxn\` source (suppressed when the matching DM conversation is on top, mirroring the \`dm/message\` suppression).
- CLAUDE.md gains \`rxn\` in the notification-sources list.

## Persistence

Reactions are in-RAM only in v1. Persisting across reboot would mean writing a small store to SD on every \`chat/reaction\` event; the value vs the wear/IO trade is marginal for what's essentially low-information ack data. Adding it later is mechanical -- the \`reactions\` table is already the canonical shape.

## Out of scope (deferred to follow-ups; the issue is wider than one reviewable change)

- **DM replies via the TXT_MSG flags byte.** The issue suggests claiming a flag bit (e.g. 0x80 = HAS_REPLY) and prefixing the text body with \`[reply_to_hash:4]\`. Wire change is small but needs its own bubble-rendering (quoted-snippet header) + scroll-to-original UX. Older receivers would see 4 bytes of garbage at the start of text -- the issue calls this out as the intentionally-lossy failure mode.
- **Channel reactions / channel replies.** Channel GRP_TXT plaintext is \`[timestamp:4][type:1][sender: text]\`; adding a reaction type means a new \`msg_type\` value plus parallel storage in the channels service. Substantial separate change.
- **\"Who reacted\" long-press detail.** \`compress()\` already returns the sender list per emoji; just needs a small list screen.

## Test plan

- [x] \`pio run\` builds clean on this branch (RAM 26.2%, Flash 58.9%).
- [ ] **Needs hardware QA** -- \`pio run -t upload\` from this worktree, then with two devices that have each other as contacts:
  - DM device A from device B. On B, open the conversation, focus the inbound bubble, press Enter, pick \"React...\", choose +1. The bubble's footer shows \`+1\` in the accent colour; the optimistic local record paints immediately.
  - On A, the corresponding bubble's footer shows \`+1\`. No toast (A is the conversation partner; we suppress when the conversation is on top).
  - Send the same reaction again from B (or the same emoji from A). Counts merge: \`+1 2\`.
  - Send a different emoji (e.g. \`OK\`) from a third device that's also in the conversation -- footer shows both pills side-by-side.
  - With the DM conversation NOT on top, react from another device. A toast appears with \"Reaction \\<emoji\\>\" / \"From \\<contact_name\\>\". Set \`notify_rxn = 0\` via the terminal and confirm the toast is suppressed.
  - Reboot one device. Confirm reactions don't survive (in-RAM only in v1, documented above).

Closes #112


xtr-bas:
Reaction transport switched from RAW_CUSTOM to a sharing-service URL ridden inside a normal DM TXT_MSG (`3f972b0`, plus a `00b2264` merge of latest `origin/test`).

**Wire change.** The 5-byte custom-packet payload is gone. Reactions are now `https://ezme.sh/#rxn/v1?h=<8-hex msg_hash>&e=<emoji_index>` sent through `dm.send()`. They inherit:
- flood routing (multi-hop reach -- RAW_CUSTOM only made it one direct hop),
- the MeshCore retry / ACK pipeline (so a missed reaction retransmits up to MAX_RETRIES, same as a DM),
- the existing ECDH+AES+HMAC envelope (no new crypto path).

**Receive side.** `services/direct_messages.lua` calls `reactions.try_handle_inbound(sender_pub, text, name)` on both decrypt paths (live RX and the `try_decrypt_pending` retroactive path) BEFORE `store_message`. When the text parses as `kind == "reaction"` the reactions store and `chat/reaction` bus event fire as before, but no visible bubble is created -- the reaction surfaces as a footer pill on the target message (or a toast when the conversation isn't on top). UX matches the previous RAW_CUSTOM behaviour.

**Changes**
- `lua/services/sharing.lua` -- new `encode_reaction(msg_hash, emoji_index)` and a `rxn/v1` branch in `parse()`. Wire-format doc block updated.
- `lua/services/reactions.lua` -- `send()` now calls `sharing.encode_reaction(...)` + `dm.send(target_pub, url)`. New `try_handle_inbound()` for the DM intercept. `custom_packets` dependency dropped; `init()` is a no-op kept for API symmetry.
- `lua/services/direct_messages.lua` -- meta-payload intercept before `store_message` on both decrypt paths.
- `lua/boot.lua` -- comment updated; reactions no longer needs to run after `custom.init()`.

`pio run` clean (RAM 27.5%, Flash 59.2%). Hardware-sensitive -- the radio path changed, so worth flashing two devices and re-running the test plan in the PR body to confirm tap-to-react, footer-update, and toast behaviour. Also worth confirming multi-hop reactions now land (the whole point of the switch).

xtr-bas:
Addressed in \`a10e3d6\`:

- **\`lua/services/reactions.lua:81\` (timestamp hash basis mismatch)** -- the actual bug was in \`direct_messages\`: \`build_inner\` reached for \`ez.system.get_time().epoch\` which doesn't exist (the binding returns \`{hour, minute, second, year, month, day}\`), so the wire timestamp was always 0; meanwhile \`dm.send\` stored \`ez.system.millis()\` for self-sent bubbles while the receive side stored \`msg_timestamp\` (epoch seconds). Both sides now use \`ez.system.get_time_unix()\`, so \`compute_msg_hash\` produces matching digests on both peers.
- **\`lua/services/reactions.lua:183\` (raw URL bubble)** -- new \`dm.send(pub, text, { meta = true })\` opt skips the local \`store_message\` + \`dm/message\` post + ACK-retry queue. Reactions ride through the same crypto + flood path but produce no bubble on the sender.
- **\`lua/services/reactions.lua:198\` (unknown-emoji silent drop)** -- \`try_handle_inbound\` now returns \`false\` for indices outside \`EMOJI_PALETTE\`, so a future palette entry from a newer peer falls through to the normal text-bubble path instead of being swallowed.
- **\`lua/boot.lua:212\` / \`CLAUDE.md:383\` (services list)** -- added \`reactions\` as item 7 in the Services boot order list and renumbered.

Hardware-sensitive: radio + DM path. Worth flashing two devices and round-tripping a reaction to verify the hash now matches on the sender's bubble. \`pio run\` clean.

PR #151, opened 2026-05-15
feat(notifications): quiet hours / Do Not Disturb schedule

xtr-bas:
## Summary

A time-window DND mode for the notifications service.

When DND is active (manual override `dnd_manual`, or schedule `dnd_enabled` with the wall clock inside `[dnd_start, dnd_end)`), incoming notifications still land in the list -- so `unread_count` advances and the user catches up on the backlog when they wake -- but they're flagged `silent`, the toast subscriber in `ezui/screen.lua` skips them, and the panel-wake-on-notification path is suppressed.

Two opt-in exemptions:
- `opts.dnd_mention` -- channel message containing the user's node name. Wired in `boot.lua`'s `channel/message` subscriber, gated on the `dnd_mentions` pref.
- `opts.dnd_fav` -- DM from a starred contact. Service hook exists; no call site yet (`services.contacts` has no favourite field, called out in CLAUDE.md as the next step).

Settings -> Notifications (new screen) toggles manual DND, the schedule, window endpoints (15-min sliders, minute-since-midnight), and the mentions exception. Window wraps midnight when `end <= start`, so 22:00..07:00 covers `[22:00, 24:00) U [00:00, 07:00)`.

DND is treated as off when the clock is unset (year < 2020) so a cold boot before NTP/GPS sync doesn't swallow notifications.

CLAUDE.md's Notifications section now documents the DND hooks + exemption flags.

Closes #116

## Test plan

- [ ] **Needs hardware QA** -- `pio run -t upload` from this worktree, then:
  - Settings -> Notifications: toggle "Manual DND now", post a synthetic notification (`-e "require('services.notifications').post({title='x',body='y',source='dm'})"`), confirm no toast renders and the panel stays dark if it was idle. Toggle off, repeat, confirm the toast shows up.
  - Set the schedule to e.g. start=now, end=now+15min, enable it, drop a fresh notification, confirm silent.
  - Enable "Allow channel mentions", have a peer send a channel message mentioning your node name during quiet hours, confirm the toast lands.
  - Confirm `unread_count()` advances even for silenced notifications (so the bar / notifications screen still shows unread).
- [x] Build is clean (`pio run` succeeds, verified locally).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #161, opened 2026-05-15
feat(chat): read receipts for DMs via ack/v1 share URI

### LINKED ISSUE #113: feat(chat): delivery acks and read receipts for DMs
## Problem

DMs go out via TXT_MSG and the sender has no idea whether the packet actually reached the contact. There's no delivery ack, no read receipt, no retry signalling -- a failed send looks identical to a successful one. On a flood-routed mesh with variable hop counts, this is the single biggest "is it broken or just slow?" UX hole.

## Proposal

Two distinct signals, both opt-in per contact (and globally togglable in Settings -> Privacy):

### Transport

Acks ride inside **TXT_MSG**, not as a `RAW_CUSTOM` packet. Two reasons:

1. **Repeaters relay them.** Stock MeshCore repeaters refuse to re-flood `RAW_CUSTOM` (see the header note in `lua/services/custom_packets.lua` and the explicit skip in `src/mesh/meshcore.cpp:201-211`). Custom packets are direct-range only, which is the exact opposite of what an ack needs -- the "is it broken or just slow?" case is multi-hop by definition.
2. **Cross-firmware graceful degradation.** The sender of a DM may be running stock MeshCore chat or any other compatible client, not ezOS. Encoding the ack as an `https://ezme.sh/#ack/...` URI inside a TXT_MSG means a non-ezOS recipient sees a readable, clickable link in their chat history instead of a mystery payload; ezOS recipients peel it off in `direct_messages.lua` and update the bubble glyph silently without rendering the URL.

URI shape (TBD, consistent with the existing `lua/services/sharing.lua` encoding):

```
https://ezme.sh/#ack/<base64url(msg_hash:4 || kind:1)>
```

`msg_hash` keeps the same convention as the reactions feature -- first 4 bytes of SHA-256 over `[src_pubkey:32][timestamp:4][text]`. `kind` is `1 = DELIVERED`, `2 = READ`.

### Delivery ack ("got it")

- Sent automatically by the receiver as soon as the TXT_MSG is decrypted successfully.
- One ack per delivered DM, deduped by hash on the sender side.

### Read receipt ("seen it")

- Sent when the user actually opens the conversation containing that message (not just when the device receives it). Same URI shape, `kind = READ`.
- Strictly opt-in -- some people don't want their reading habits broadcast. Default off. Per-contact override.

### Sender-side state machine

Local message record gains a `delivery_state` field: `pending | sent | delivered | read | failed`. Renders as a 1-glyph footer on the bubble (clock / check / double-check / eye / "!" with a "tap to retry" action). Failed = no ack received within `T_FAIL` seconds (start with 60s; tune from radio data).

### Retry

When the state crosses to `failed`, expose a "Resend" entry in the bubble context menu. Don't retry automatically -- the user knows context the device doesn't (maybe they typed it on the wrong contact).

## Airtime cost

TXT_MSG wraps the payload in the X25519/AES/HMAC envelope (~20 bytes overhead plus padding to a 16-byte boundary), so an ack URI is materially more expensive on-air than a 5-byte `RAW_CUSTOM` would have been. Read receipts staying default-off mitigates this; we may also want to coalesce acks if multiple DMs from the same peer land within a short window.

## Privacy / security

- Acks inherit TXT_MSG's per-pair X25519/AES/HMAC envelope, so a passive observer learns "an ack-shaped TXT_MSG happened" but nothing about which message it acknowledges -- the 4-byte hash is inside the ciphertext.
- Read receipts can be disabled per-contact AND globally; the global toggle defaults to ON for delivery, OFF for read.

## Out of scope

- Per-channel "seen by N" UI (channels don't have a tight enough recipient set for this to mean much).
- Typing indicators (much more airtime, much less value on a mesh).

## Files

- `lua/services/sharing.lua` -- new `ezme.sh/#ack/...` URI encode/decode helpers.
- `lua/services/direct_messages.lua` -- per-message state machine, ack URI emission, inbound ack URI detection (peel before rendering), retry hook.
- `lua/screens/chat/*` -- bubble footer glyph + retry action.
- `lua/screens/settings/privacy.lua` (new or extend existing) -- global toggles.

xtr-bas:
## Summary

Adds the missing half of issue #113. The MeshCore protocol-level ACK already covers \"delivered\" (existing status flips \`pending -> delivered\` when the ACK lands); this PR adds the **read receipt** signal that fires when the user actually opens a DM conversation containing an unread bubble.

## Wire format

A read receipt rides inside a normal TXT_MSG, encoded as the new ack share URI -- so cross-firmware chat clients see a clickable link rather than mystery bytes; ezOS recipients peel it off without rendering the URL as a bubble.

\`\`\`
https://ezme.sh/#ack/v1?t=<base64url(msg_hash:4 || kind:1)>
kind: 1 = DELIVERED   (constant exists; ezOS doesn't emit -- protocol ACK already does this)
      2 = READ
\`\`\`

\`msg_hash\` matches the per-message tag convention from issue #112: first 4 bytes of \`sha256([sender_pubkey:32][timestamp:4 LE][text])\`. Both peers reproduce it from their own copy of the message; no extra wire bits required.

## Components

- **\`services/sharing\`**: new \`encode_ack(msg_hash, kind)\` builder + \`ACK_VERB\` branch in \`parse()\`. Constants \`sharing.ACK_DELIVERED\` / \`sharing.ACK_READ\`.
- **\`services/direct_messages\`**:
  - \`mark_read(peer)\` also calls \`maybe_send_read_receipts(peer)\` (gated on the \`send_read_rcpt\` pref, default \"0\"). Walks the conversation, computes per-message hash, sends one ack URI per receipt-eligible inbound bubble through the existing \`dm.send\` pipeline. An in-RAM dedup cache prevents re-emission when the user re-enters the same conversation.
  - Receive path peels ack URIs: an inbound TXT_MSG whose text parses as a \`kind=READ\` ack is matched against the local self-sent bubbles by \`msg_hash\` and flips that bubble's status to \"read\". The ack is NOT stored as a chat row. The MeshCore protocol ACK still fires so the sender's \`pending_acks\` entry clears.
- **\`screens/chat/chat_common\`**: status dot for \`status == \"read\"\` paints the accent colour with an outer ring -- the \"double check\" pattern from the issue, adapted for a 3 px dot.
- **\`screens/settings/privacy_settings\`** (new): global \"Send DM read receipts\" toggle wired to the \`send_read_rcpt\` pref. Notes that delivery acks are always on and not user-tunable.
- **\`screens/menu\`**: new \"Privacy\" row in the System tab.

## Out of scope (deliberately deferred)

- **Per-contact override for read receipts.** Would live on the contact's detail screen alongside notification mute; v1 covers the global default only.
- **DELIVERED kind in the ACK URI.** The constant exists for symmetry / cross-firmware compatibility but ezOS senders don't emit it -- the existing MeshCore protocol ACK already covers delivery, and double-emitting would waste airtime.
- **T_FAIL auto-transition** for messages never ack'd. The existing \`pending_acks\` layer already exposes the \"unconfirmed\" state on timeout; the issue's T_FAIL is a tuning knob on that path, not a new state.

## Test plan

- [x] \`pio run\` builds clean on this branch (Flash 58.9%).
- [ ] **Needs hardware QA** -- \`pio run -t upload\` from this worktree on two devices that have each other as contacts:
  - On device A, turn on Settings -> Privacy -> Send DM read receipts.
  - On device B, send a DM to A. On A, observe the unread badge.
  - On B, the sent bubble shows the green \"delivered\" dot (existing behaviour) within seconds.
  - On A, open the conversation. The unread counter clears; behind the scenes A emits an ack/v1 URI back to B.
  - On B, the sent bubble's dot flips to the accent-coloured ringed mark (\"read\").
  - Leave A's conversation, send another DM from B, open it on A again. Confirm only the new message triggers a fresh receipt; the previously-acked one isn't re-emitted (dedup cache).
  - Turn the toggle off on A. Re-test: opening the conversation no longer emits the URI; B's bubble stays at \"delivered\".
  - Confirm a third device (or simulated peer) that doesn't speak ezOS sees the ack URI as a normal clickable chat bubble (i.e. the link renders rather than being silently dropped by the protocol layer).

Closes #113


xtr-bas:
Merged \`origin/test\` in \`2912bd0\`. \`lua/screens/menu.lua\` conflicted because \`test\`'s \`feat(notifications): quiet hours / Do Not Disturb schedule\` (#151) replaced the same Settings list slot this PR uses. Resolved by keeping both rows (Notifications above Privacy); both screens remain reachable from Settings. \`pio run\` clean.

xtr-bas:
Addressed the inline review feedback in \`469ee63\` (preceded by a \`test\` merge in \`1b204b2\`):

- **\`lua/services/direct_messages.lua:818\`** (timestamp domain mismatch) -- \`dm.send\` now stashes the wire-format epoch (\`ez.system.get_time().epoch\`, same value \`build_inner\` packs into the encrypted plaintext) as \`wire_ts\` alongside the existing millis-based \`timestamp\`. The ACK match loop hashes against \`m.wire_ts\` (with a \`m.timestamp\` fallback for bubbles stored before this fix landed) so the digest agrees with what the receiver computed. \`timestamp\` is kept for display/ordering/dedup.

\`pio run\` clean. Hardware-sensitive (DM wire path + ACK loop); please flash two devices and round-trip a DM to verify the accent-coloured read ring lights up on the sender side after the receiver opens the conversation.

xtr-bas:
Addressed both inline asks in \`53757ac\` (preceded by a \`test\` merge in \`65cf939\`):

- **\`lua/services/direct_messages.lua:1224\`** (ack URL stored as local bubble) -- \`maybe_send_read_receipts\` now passes \`{ protocol = "ack" }\` and \`dm.send\` short-circuits the storage + bus-post path when \`opts.protocol == "ack"\`. The ack still rides the same crypto / flood / retry plumbing but produces no \`is_self\` bubble in the sender's conversation.
- **\`lua/services/direct_messages.lua:1188\`** (in-memory-only dedup re-floods after reboot) -- replaced the \`_G._dm_read_receipts_sent\` cache with a persisted \`msg.receipt_sent\` flag on each inbound entry. \`do_save\` serialises it alongside the other per-message fields; \`maybe_send_read_receipts\` checks it before sending and stamps + \`schedule_save()\`s on first send. Reboots no longer trigger the historical re-flood; repeated \`mark_read\` calls still skip cheaply.

\`pio run\` clean. Hardware-sensitive: LoRa send path + DM persistence schema. Please flash two devices and round-trip a DM with \`send_read_rcpt = "1"\`, then reboot the reader and re-open the conversation to confirm zero extra ack TXT_MSGs go out.

xtr-bas:
Addressed all four new inline asks in \`2ed7370\` (\`test\` was already up to date):

- **\`lua/services/direct_messages.lua:1085\`** (wire_ts will not match the embedded plaintext timestamp) -- \`build_inner\` now takes an optional \`wire_ts\` argument and uses it as the embedded timestamp; \`transmit\` threads it through; \`dm.send\` passes the value it captured at call time. The retry path stashes \`wire_ts\` on the \`pending_acks\` entry and re-uses it so retry attempts carry the same timestamp. Bubble and wire agree by construction now.
- **\`lua/services/direct_messages.lua:1240\`** (\`receipt_sent\` stamped before send confirmed) -- added \`opts.on_success\` to \`dm.send\`, fired in-spawn after \`transmit\` confirms the packet was queued (\`pcall\`-guarded). \`maybe_send_read_receipts\` stamps \`msg.receipt_sent\` + \`schedule_save\` inside that callback, so a cold-boot \`mark_read\` against a not-yet-initialized mesh no longer burns the dedup slot.
- **\`lua/screens/settings/privacy_settings.lua:41\`** (direct state mutation without set_state) -- \`on_change\` now writes the pref first, then calls \`self:set_state({ read_receipts = val })\` so the canonical state stays in sync with the toggle widget.
- **\`lua/screens/settings/privacy_settings.lua:1\`** (no manual doc page) -- added \`lua/docs/manual/settings/privacy.md\` describing the read-receipt toggle, the LoRa airtime cost, why it defaults to off, and how it relates to MeshCore's always-on delivery ack.

\`pio run\` clean. Hardware-sensitive (DM crypto + radio path + persistence schema); please flash two devices and round-trip a DM with the receipt toggle on, then reboot the reader and confirm: no extra ack TXT_MSGs on re-open, the sender's bubble flips to "read" (the ts mismatch should be gone), and the dedup slot survives a cold-boot \`mark_read\`.

PR #162, opened 2026-05-15
feat(ui): cross-cutting Search screen

### LINKED ISSUE #114: feat(ui): mesh-wide search across contacts, channels, and messages
## Problem

There is no cross-cutting "find" path on the device. To locate a contact you scroll the contacts screen; to find a message you scroll the relevant chat; to find a channel you remember its name. On a device with months of history and dozens of contacts that's painful.

## Proposal

A single Search screen reachable from the desktop (and from a `/` keyboard shortcut where the focus is a list).

### Index sources

- **Contacts:** name, optional note, last-seen pubkey hash prefix.
- **Channels:** channel name, last message preview.
- **Messages:** DM and channel text. Plain substring match against the stored history -- no fulltext index. On a T-Deck with ~MB of history a linear scan is fine; we can add a trigram index later if it becomes slow.
- **Apps / settings panels:** so the user can find "brightness" without remembering which Settings subscreen owns it. Static registry, loaded once.

### UI

- Single input field at the top, results grouped by source ("Contacts", "Channels", "Messages", "Settings"). Each group caps at the first 10 matches with a "More..." entry.
- Tap a result -> jump to the right screen (contact details, channel history scrolled to the matching message, etc.).
- Case-insensitive, ASCII-fold first (printable ASCII is the safe surface per CLAUDE.md fonts rule).

### Performance

- Lazy: don't build any index until the screen opens. The message scan is bounded by total stored history, which is finite and already in RAM/SD.
- Debounce keystrokes at 150 ms so we're not re-scanning per char.
- Yield in a coroutine every N rows so the UI stays responsive.

## Out of scope

- Search inside markdown docs (help has its own browser).
- Regex / fuzzy match. Substring is enough for v1.

## Files

- `lua/screens/tools/search.lua` (new).
- `lua/screens/desktop.lua` -- add icon.
- `lua/services/{contacts,channels,direct_messages}.lua` -- expose a `search(query)` helper each.
- Settings registry -> add a flat panel-name list.

xtr-bas:
## Summary

Single Search screen reachable from More -> Tools -> Search. Input field at the top accepts a substring; matches land grouped by source (Contacts, Channels, Messages-DM, Messages-channel, Settings). Each group caps at 10 rows with a \"More...\" sentinel when truncated -- refining the query is the recovery path, no pagination.

## How it works

- All scans run against the in-RAM state already maintained by \`services.contacts\`, \`services.channels\`, and \`services.direct_messages\`. No new index; on a T-Deck with the typical history size a linear walk is fast enough.
- Debounced at 150 ms via \`Search:update()\` comparing the last keystroke time against the last query that produced results; the actual scan runs inside \`spawn()\` so the UI never blocks on long history. Queries the user typed past while the previous scan was in flight are discarded on result-handoff.
- Open actions jump to the natural screen: contact -> DM conversation, channel -> channel chat, DM/channel message -> the conversation containing it, settings panel -> the panel.

## Where Search lives

The desktop has a fixed 4-icon layout (Messages / Contacts / Map / More); threading a 5th in would mean a layout change. Search goes under More -> Tools for now, alongside Notifications / Help / Terminal. The issue also suggested a \`/\` keyboard shortcut from list screens, but \`/\` isn't on the T-Deck's physical keyboard (it's an Alt+letter chord), so I dropped that part rather than design a chord whose name isn't \"/\".

## Settings registry

A hand-maintained static list at the top of \`search.lua\` (\`SETTINGS_PANELS\`). Adding a new searchable panel means adding a row here. The subtitle field is included in the hay so a query like \"brightness\" still finds Display via its subtitle.

## Out of scope (deferred)

- **Auto-scroll to the specific matching message** inside a chat. The open action lands on the conversation but the user still has to scroll. A scroll-to-id would need each chat screen to take an optional \`scroll_to_msg_id\` initial state.
- **Regex / fuzzy match.** Substring is enough for v1, per the issue.
- **Search across markdown docs.** The help screen has its own browser.
- **\`/\` keyboard shortcut.** No \"/\" key on-device.

## Test plan

- [x] \`pio run\` builds clean on this branch.
- [x] \`luac32 -p\` passes.
- [ ] **Needs hardware QA** -- \`pio run -t upload\` from this worktree:
  - Open More -> Tools -> Search. Input is focused for immediate typing.
  - Type a contact name fragment; matches appear under \"Contacts\". Tap one -> DM conversation opens.
  - Type a channel name. Both the channel itself and any messages containing the substring appear under separate groups.
  - Type a substring of a recent message body. The message appears under \"Messages (DM)\" or \"Messages (channel)\" with the surrounding text snippet. Tap -> the conversation opens.
  - Type \"brightness\". Settings -> Display appears under \"Settings\" via the subtitle hit.
  - Type a query with > 10 matching messages in one channel: the group caps at 10 and shows \"More...\". Refining the query shrinks the list as expected.
  - Type fast and confirm there's no per-keystroke lag (debounced).

Closes #114


xtr-bas:
Addressed in \`18e47dc\`:

- **\`lua/screens/tools/search.lua:295\` (stale-query guard no-op)** -- snapshot \`s.query\` into a value local \`q\` before \`spawn()\`. The old check compared the same table reference both before and after the yield, so the guard always passed.
- **\`lua/screens/tools/search.lua:129\` (peer-data not sanitized)** -- added a local \`ascii_safe()\` mirroring the notifications service's policy and wrapped every peer-originated string assembled into \`ui.list_item\` titles/subtitles: contact name, channel name, DM sender, channel sender, and the message-text snippets. Bytes outside \`0x20..0x7E\` now substitute \`?\` instead of rendering as \`[]\` glyph boxes.

\`2edf279\` is just \`origin/test\` merged in (CLEAN auto-merge); \`pio run\` clean.

xtr-bas:
Addressed both outstanding inline comments in \`ab1f709\` (preceded by a clean \`test\` merge in \`22f6a1f\`); the third comment (peer-data sanitization on DM/channel rows) is already covered by the previous commit \`18e47dc\` -- \`search.lua\` lines 135, 190, 222 all run through \`ascii_safe()\`.

- **\`lua/screens/tools/search.lua:370\`** (no keyboard back-path) -- \`lua/ezui/widgets.lua\`'s \`text_input\` gains an \`on_empty_backspace\` hook; it fires when BACKSPACE arrives at cursor 0 and the caller provided it (backwards-compatible for every other call site). \`search.lua\` wires it to \`screen_mod.pop()\`, so on a T-Deck (no Esc key, focus.handle_key routes BACKSPACE to the input while editing) the user gets a keyboard back-path out of the auto-edit screen.
- **\`lua/screens/tools/search.lua:1\`** (missing manual entry) -- added \`lua/docs/manual/search.md\` describing the Search screen and listed it in \`lua/docs/manual/index.md\` under Sections.

\`pio run\` clean. UI-sensitive (keyboard input chokepoint); please flash and re-test that BACKSPACE-to-back works on an empty query AND that typing then BACKSPACE-to-delete still works in the input field.

PR #165, opened 2026-05-15
feat(power): battery-aware radio + display duty cycle

### LINKED ISSUE #117: feat(power): battery-aware radio duty cycle
## Problem

The radio, GPS, and periodic background syncs all run at full cadence regardless of battery state. A user at 5% has the same airtime / power profile as one at 100%. A "low power" mode that backs off non-essential traffic would meaningfully extend uptime in the field.

## Proposal

A power-policy module that adjusts cadence based on battery percentage.

### Tiers

| Battery | Mode | Behaviour |
|---|---|---|
| > 30% | Normal | Current behaviour |
| 10-30% | Frugal | ADVERT cadence halved, hourly GPS sync skipped, NTP sync every 6h instead of hourly, screensaver dim faster |
| < 10% | Survival | ADVERT cadence quartered, GPS sync disabled, NTP disabled, all non-DM custom packets suppressed, display brightness clamped to 30%, radio TX power dropped one notch |

Transitions are sticky -- once Survival kicks in, hysteresis keeps it there until battery climbs back above 15% (avoid flapping at the boundary).

### Settings

Settings -> Power -> "Low power thresholds":
- Override which mode applies at which percentage.
- "Always Normal" toggle for users on a powered desk.
- "Force Survival now" manual switch (useful for "I need this to last another 4 hours").

### Surfacing

- Status bar shows a small "low power" glyph in Frugal, a more prominent one in Survival.
- First entry into Frugal/Survival posts a one-shot notification ("Low power mode -- reduced radio activity") so the user isn't surprised by missed traffic.

### Hooks

- Mesh module: expose `set_advert_period(s)` / `set_tx_power(dbm)`.
- GPS service: respect `power.gps_allowed()` in the sync loop.
- Notifications: tag DND-like suppression source so the user can opt out.

## Out of scope

- Per-app power profiles.
- USB-detect to auto-disable (could be a follow-up; needs a hardware check).

## Files

- `lua/services/power.lua` (new) -- tier evaluator + bus events.
- `lua/services/{gps,ntp,custom_packets}.lua` -- consult policy.
- `src/lua/bindings/mesh_bindings.cpp` -- expose advert period / TX power knobs if not already.
- `lua/screens/settings/power.lua` (new or extend battery panel).

xtr-bas:
## Summary

A `services/power` policy that backs off radio, GPS, NTP, display
brightness, and non-DM custom packets as the battery drains. Three
tiers with hysteresis:

| Battery   | Tier      | Side effects                                                                          |
|-----------|-----------|---------------------------------------------------------------------------------------|
| > 30 %    | Normal    | Default. ADVERT 120 s, TX 22 dBm.                                                     |
| 10 - 30 % | Frugal    | ADVERT 240 s, GPS sync paused. NTP unchanged.                                         |
| < 10 %    | Survival  | ADVERT 480 s, TX 17 dBm, GPS sync stopped, NTP stopped, non-DM RAW_CUSTOM blocked, display clamped to <= 80/255. |

Hysteresis: Frugal leaves at >= 35 %, Survival at >= 15 %. Charging
snaps the device toward Normal regardless of the reported level.

DM TXT_MSG traffic is **never** gated -- someone in trouble might be
on 4 % battery and we still need their messages to land.

## What's wired

- New `lua/services/power.lua` with the tier truth table, `evaluate()`
  loop, `current_mode()` / predicates (`gps_sync_allowed`,
  `ntp_allowed`, `allow_non_dm`), and `power.start()` polling every
  30 s. Emits `power/mode_changed` on the bus.
- `lua/services/gps.lua` -- `try_sync_once` early-returns when
  `power.gps_sync_allowed()` is false.
- `lua/services/custom_packets.lua` -- `M.send()` returns false when
  `power.allow_non_dm()` is false.
- `lua/services/notifications.lua` is the recipient of one-shot
  transition toasts (source `"power"`, `dismiss_source` first).
- `lua/boot.lua` -- starts the service after gps/ntp/custom_packets.
- `lua/ezui/screen.lua` + `widgets.lua` -- new `power_tag` field on
  `screen.status`, rendered as `lp` / `LP` in `ACCENT` left of the
  battery in the status bar.
- New `lua/screens/settings/power_settings.lua` reached from
  Settings -> Power. Two mutually-exclusive toggles for "Always
  Normal" / "Force Survival now" backed by `pwr_always_norm` /
  `pwr_force_surv` (both <= 15 chars).
- New user-manual page `lua/docs/manual/settings/power.md`.
- CLAUDE.md service list + a new "Power policy" section under the UI
  architecture chapter.

## Out of scope for this PR

The issue mentions threshold customization (override which percentage
maps to which tier). Defaults are baked in for v1 -- the manual
override toggles cover the "I disagree with the schedule" cases. Easy
to add later.

The issue also mentions "NTP every 6h" in Frugal. lwIP's SNTP cadence
isn't exposed at the Lua surface; this PR stops NTP cleanly in
Survival but leaves Frugal unchanged. Reaching into lwIP for a custom
cadence would need a new C++ binding -- worth doing only if the
"every 6 h" target turns out to be load-bearing.

## Test plan

- [x] `pio run` builds clean on this branch.
- [ ] **Needs hardware QA** -- flash. Battery in Settings should show
      normal. Pull the USB cable and let the device run to < 30 %.
      Confirm a "Low power mode" toast fires, the status bar shows
      "lp" left of the battery, and ADVERT cadence in serial logs
      doubles.
- [ ] **Needs hardware QA** -- continue down to < 10 %. Confirm a
      "Survival mode" toast fires, the bar shows "LP", brightness
      clamps to a dimmer level, and `ez.radio.get_settings()` shows
      `tx_power = 17`.
- [ ] **Needs hardware QA** -- plug USB back in. Status bar tag
      should clear, brightness restore to the pref value, and
      `ez.mesh.get_announce_interval()` should drop back to 120000.
- [ ] **Needs hardware QA** -- Settings -> Power -> "Force Survival
      now" should force the tier regardless of battery. Toggling
      "Always Normal" should clear the Force Survival pref bit so
      we never end up with both on.
- [ ] **Needs hardware QA** -- send a DM in Survival; confirm it
      goes out (the gate is only for RAW_CUSTOM, not TXT_MSG).

Closes #117


xtr-bas:
Addressed in \`dbcda0a\`:

- **\`lua/screens/settings/power_settings.lua:66\` + \`:86\` (Lua ternary)** -- replaced \`v and false or state.x\` with \`not v and state.x\`. The old form always returned \`state.x\` (because \`v and false\` evaluates to \`false\` for any \`v\`, then \`false or x = x\`), so flipping one toggle on left the mutually-exclusive partner visually ON until a rebuild.
- **\`lua/services/power.lua:198\` (false low-battery toast on plug-in)** -- threaded \`charging\` through \`notify_transition\` and bail out early for charge-induced step-ups. A survival -> frugal transition triggered by plugging in no longer fires a "Battery low" toast; the next 30 s tick promotes to normal cleanly.

Hardware-sensitive: power policy + display brightness. \`pio run\` clean.

PR #166, opened 2026-05-15
feat(security): encrypt identity private key at rest behind passphrase

### LINKED ISSUE #118: feat(security): encrypted at-rest storage for identity keys
## Problem

The Ed25519 identity keypair lives in NVS in plaintext (`privkey`, `pubkey`). A device that's lost, stolen, or briefly out of sight can be read out over USB or by reflashing a permissive firmware -- and that key IS the user's identity on the mesh. Anyone with the bytes can impersonate them, decrypt their DM history with any contact, and join their encrypted channels.

## Proposal

Wrap the private key with a passphrase-derived KEK.

### Format

- Argon2id (or PBKDF2-SHA256 if Argon2 won't fit in flash; we have NaCl-style primitives already in `ez.crypto`) over a user passphrase + 16-byte random salt -> 32-byte KEK.
- KEK wraps the 64-byte privkey via XSalsa20-Poly1305 (we have it from the NaCl binding family). Output: 16-byte salt + 24-byte nonce + 80-byte ciphertext = 120 bytes. Replaces the current `privkey` NVS entry.
- KDF cost picked so unlock takes ~1-2s on the ESP32-S3 (slow enough to deter brute-force on extracted bytes, fast enough for a usable lockscreen).

### Boot flow

1. Boot detects wrapped key -> shows a passphrase entry screen before any mesh init.
2. On unlock success the plaintext privkey is held in RAM only (zeroed on shutdown / panic).
3. On unlock failure, exponential backoff (1s, 2s, 4s, ... up to 60s) to slow on-device guessing.

### Migrations

- First boot after the update detects an unwrapped key, prompts the user to set a passphrase (offering "skip / leave unencrypted" for users who explicitly opt out -- some will, and we shouldn't force it).
- A "Change passphrase" entry in Settings -> Security re-wraps with a new salt + KEK.
- An "Remove passphrase" entry restores the plaintext entry (with a strong confirmation prompt).

### Lockscreen tie-in

This pairs naturally with the device-lock issue: the passphrase used to unwrap the key can also gate the lockscreen, so the cost is paid once per boot rather than per unlock.

## Out of scope

- TPM-style hardware-backed key storage (the ESP32-S3 has eFuse but no general-purpose enclave; complicates recovery).
- Encrypting message history. Bigger change; address after the identity story is solid.

## Files

- `src/lua/bindings/crypto_bindings.cpp` -- expose Argon2id / pbkdf2 + XSalsa20-Poly1305 if not already.
- `src/main.cpp` or boot path -- gate mesh init on unlock.
- `lua/screens/onboarding/passphrase.lua` (new) -- set / change / unlock flows.
- `lua/services/migrations.lua` -- one-shot migration from plaintext to wrapped.

xtr-bas:
## Summary

Encrypt-at-rest for the Ed25519 identity private key. Default
behaviour is unchanged -- when no `id_wrap` blob exists, boot, sign,
send all work exactly as today. Settings -> Security -> Set
passphrase opts in.

## Threat model

Plaintext `privkey` is readable over USB / serial / reflash by
anyone with physical access. Wrapping protects against an attacker
who has the device but not the passphrase. Once unlocked in RAM,
the terminal can call `ez.identity.*`; that's outside the wrap's
threat model (CLAUDE.md already documents physical-access-with-
screen-unlocked as "game over").

## Crypto

- KEK: PBKDF2-SHA256 (100 000 iterations, ~1 s on ESP32-S3) over
  passphrase + 16-byte random salt -> 32 bytes.
- Wrap: AES-256-GCM (12-byte random nonce, 16-byte tag). The
  32-byte public key is bound into the AEAD's AAD so the wrapped
  blob is tied to this device's identity.
- Wrap format (binary, `EZL1` magic, see `services/identity_lock`
  header for the byte layout): magic, version, kdf id, iters,
  salt_len + salt, nonce_len + nonce, ct_len + ct.

## What's wired

**C++ side:**
- 3 new `ez.crypto.*` bindings (PBKDF2, AES-GCM encrypt/decrypt)
  using mbedTLS.
- `Identity::init()` detects `id_wrap` and enters `_locked` mode:
  pubkey + name load, `_hasKeypair=false` so sign / send_announce
  refuse cleanly.
- New `Identity::unlock(priv, pub, name)` + NVS helpers exposed as
  `ez.identity.*`. The `get_privkey_for_wrap` binding is a one-way
  trapdoor: it refuses once a wrapped blob exists, so an unlocked
  device's terminal cannot read the cleartext key.
- Inconsistent-state recovery: if both `id_wrap` and plain
  `privkey` are in NVS (mid-wrap power loss), the next boot
  deletes the wrapped blob and boots unwrapped on the original key.

**Lua side:**
- `services/identity_lock` owns the wrap format and the
  wrap / unwrap / remove / change ceremonies. Every destructive
  step is gated on a successful decrypt self-test so a buggy KDF /
  cipher / blob encoder cannot lock the user out.
- `screens/onboarding/passphrase` is the unified set / unlock /
  change / remove screen, with exponential backoff (1, 2, 4, 8,
  16, 32, 60 s) on unlock failure.
- `screens/settings/security` surfaces Set / Change / Remove
  depending on current state. Reachable from Settings.
- `boot.lua` pushes the unlock screen at the end of boot when
  `ez.identity.is_locked()`; services have already started but are
  harmlessly inert. After unlock the mesh layer regains signing
  capability with no service restart, and an immediate fresh
  ADVERT goes out.

## Test plan

- [x] `pio run` builds clean.
- [ ] **Needs hardware QA -- unencrypted device unchanged.** Flash
      to a device with the existing plaintext `privkey`; confirm
      boot, sign, send all work as before. Settings -> Security
      should show "Set passphrase" only.
- [ ] **Needs hardware QA -- wrap.** Settings -> Security -> Set
      passphrase, enter "test" twice. Wait for the self-test +
      delete, reboot. Boot should land on the unlock screen.
- [ ] **Needs hardware QA -- correct unlock.** Type "test", confirm
      the screen pops and the desktop appears. Confirm ADVERTs go
      out (serial log).
- [ ] **Needs hardware QA -- wrong unlock.** Type "wrong", confirm
      the backoff timer kicks in and increases on repeated wrong
      attempts.
- [ ] **Needs hardware QA -- change passphrase.** From an unlocked
      device, Settings -> Security -> Change passphrase, enter
      current = "test", new = "test2" twice. Reboot, confirm only
      "test2" unlocks.
- [ ] **Needs hardware QA -- remove passphrase.** From an unlocked
      device, Settings -> Security -> Remove passphrase, enter
      "test2". Reboot, confirm no unlock prompt and the existing
      identity still works.
- [ ] **Needs hardware QA -- inconsistent-state recovery.** Manual:
      from an unlocked device, write a synthetic `id_wrap` blob
      via the terminal but DON'T delete `privkey`. Reboot. The
      C++ side should log "Inconsistent state (plain + wrapped);
      clearing wrap" and boot normally.

## Out of scope

- Encrypting message history (the issue calls this out as
  out-of-scope; bigger change, address after identity).
- TPM-style hardware-backed storage (ESP32-S3 has no general-
  purpose enclave).
- Argon2id was mentioned in the issue as the preferred KDF; mbedTLS
  on the ESP32 framework doesn't ship Argon2id without pulling in
  a separate library. PBKDF2-SHA256 at 100 000 iters is the
  cost-equivalent compromise.

Closes #118


xtr-bas:
Addressed all four inline comments in `acaa614` (preceded by a `test` merge in `58394bd`):

- **`src/mesh/identity.h:78`** -- `saveNodeNameOnly()` no longer calls `saveToNVS()`; it now opens NVS directly and writes only `KEY_NODE_NAME`, so calling it on an unlocked-from-wrapped device can't resurrect `KEY_PRIVATE_KEY` and trigger the inconsistent-state rollback on the next boot. Body moved to `identity.cpp` to avoid pulling `<Preferences.h>` into the header.
- **`lua/services/identity_lock.lua:176`** -- both `wrap()` and `change()` now run the self-test through `read_wrapped_blob()` + `decode_blob()` + the full decrypt path, mirroring what `lock.unwrap` does at boot. `change()` also stashes the old wrapped blob before overwriting so a self-test failure can roll back to the user's existing passphrase instead of leaving them stuck on a broken new blob.
- **`lua/screens/onboarding/passphrase.lua:225`** -- the `set` / `change` / `remove` failure branch now passes `pass = ""`, `confirm = ""`, `old_pass = ""` to `set_state` so a walked-away device doesn't leave the current passphrase visible to a shoulder-surfer.
- **`lua/screens/menu.lua:163`** -- added a Security section to `lua/docs/manual/settings.md` linking through to the existing `settings/security.md` sub-page so the on-device manual reader surfaces it alongside Display / Firmware / etc.

`pio run` clean. Hardware-sensitive surface (boot-time identity unlock, passphrase entry on the T-Deck keyboard); please flash and re-test the wrap / unwrap / change / remove flows before merging.

PR #167, opened 2026-05-15
feat(security): PIN / passphrase session lockscreen

### LINKED ISSUE #119: feat(security): PIN / passphrase lockscreen
## Problem

The device has no lock. Anyone who picks it up reads the full chat history, can send messages as you, and can extract identity keys. The screensaver dims the display but doesn't gate input.

## Proposal

A lockscreen between the screensaver and the desktop.

### Lock modes (Settings -> Security)

- **None** (default, current behaviour).
- **PIN** -- 4-8 digit numeric. Easier to enter on the T-Deck QWERTY than a passphrase.
- **Passphrase** -- any ASCII string. Pairs with the at-rest key encryption issue so a single secret unlocks both.

### Lock triggers

- After screensaver activates AND screen-off has been active for >= N seconds (config: 30s default).
- Manually via Alt+L from any screen (CLAUDE.md keyboard rule: no Ctrl).
- On boot, if a lock mode is set.

### Unlock UI

- Single-screen prompt. PIN screen shows dots, passphrase screen shows the standard text input.
- Exponential backoff on failed attempts -- 1s, 2s, 4s ... capped at 60s. Persisted across reboots (write the failure count to NVS so power-cycling doesn't reset it).
- After 10 consecutive failures, wipe-on-fail (optional, off by default, lives in Settings).

### Notification behaviour while locked

- Toasts still render with title only ("Message from <name>") unless "hide preview" is on, in which case just "New message".
- DND interacts normally.

### Trackball / touch handling

- Lockscreen disables all touch / trackball except the input field. Wake-event guard applies (CLAUDE.md "Touch input and the screensaver wake gate") -- a wake-tap from the screensaver should NOT inadvertently click a digit on the PIN screen.

## Out of scope

- Biometric unlock (no hardware).
- Per-app lock.

## Files

- `lua/screens/lockscreen.lua` (new).
- `lua/services/screensaver.lua` (or wherever it lives now) -- transition to lockscreen instead of straight to desktop.
- `lua/screens/settings/security.lua` (new) -- mode + timer + wipe-on-fail.
- NVS keys: `lock_mode`, `lock_secret_hash`, `lock_fail_count`, `lock_timeout_s`.

xtr-bas:
## Summary

A session-level lock between the screensaver and the desktop. While
the device is locked, keyboard / trackball / touch are intercepted
by an unlock prompt; nothing under the prompt receives keypresses.

Three modes (Settings -> Lockscreen):
- **Off** (default).
- **PIN** -- 4-8 digits. Easier to type on the T-Deck (digits are
  alt + the letter row).
- **Passphrase** -- any ASCII string.

Lock triggers:
- Every boot when a mode is configured.
- When the screensaver activates (the lock sits under the screensaver
  overlay so a wake-tap lands you in the prompt, not on the desktop).
- **Alt+L** from any screen.

Failure backoff: 1 s, 2 s, 4 s, 8 s, 16 s, 32 s, 60 s, persisted to
NVS so a power cycle resets the wait but not the count.

## What's wired

- `lua/services/lockscreen.lua` -- mode + secret + fail-count
  storage, `lock()` / `try_unlock()` / `setup()` / `clear()`. Uses
  a Lua-side PBKDF2-SHA256 over `ez.crypto.hmac_sha256` (50 000
  iters + 16-byte random salt) so the service is self-contained
  and doesn't depend on the new `pbkdf2_sha256` binding from
  #118's branch. Constant-time hash compare on verify.
- `lua/screens/tools/lockscreen.lua` -- the unlock overlay. Mode-
  aware UI (dots for PIN, plain text for passphrase). BACKSPACE
  and ESCAPE are swallowed -- no exit path.
- `lua/screens/settings/lockscreen_settings.lua` -- mode picker,
  secret + confirm fields, Apply button. Reachable from Settings.
- `lua/boot.lua` -- subscribes to `lockscreen/locked` and pushes
  the overlay. Boot-time `lock.lock()` arms the prompt when a
  mode is set.
- `lua/ezui/screen.lua` -- two integrations: (a) Stage 2 idle
  ladder transition calls `lockscreen.maybe_lock("idle")` so the
  screensaver-fire arms the lock; (b) Alt+L global chord locks
  immediately.
- User-manual page at `lua/docs/manual/settings/lockscreen.md`.

## What is NOT in this PR

- "Wipe on N failures" is mentioned in the issue as optional /
  off-by-default. Skipped for v1 to keep the PR tight; easy to
  bolt onto the existing fail counter.
- This lock does not encrypt anything at rest. Combine with
  #118's Settings -> Security wrap for at-rest protection.

## Test plan

- [x] `pio run` builds clean.
- [ ] **Needs hardware QA -- baseline.** With mode = Off, confirm
      the device boots and behaves exactly as today.
- [ ] **Needs hardware QA -- set PIN.** Settings -> Lockscreen ->
      mode PIN, "1234" twice, Apply. Idle out until the screensaver
      activates; wake. Confirm the lockscreen prompt is visible.
- [ ] **Needs hardware QA -- correct unlock.** Type 1234 (= alt+Q
      alt+W alt+E alt+R), Unlock. Confirm the screen pops and you
      land on whichever screen was active before lock.
- [ ] **Needs hardware QA -- wrong unlock.** Type "9999". Confirm
      the cooldown message appears with the next-attempt wait,
      and that successive wrong attempts increase the wait.
- [ ] **Needs hardware QA -- Alt+L.** From any screen with mode
      set, press Alt+L. Lock should fire immediately.
- [ ] **Needs hardware QA -- boot lock.** Reboot. Confirm the
      lock prompt is the first screen.
- [ ] **Needs hardware QA -- power-cycle cooldown.** After enough
      wrong attempts to be on a long cooldown, reboot. Confirm
      the wait is reset (you can retry immediately) but the
      fail count is preserved (next wrong attempt jumps back to
      the long wait).

Closes #119


xtr-bas:
Addressed both inline comments in \`01200e9\` (preceded by a \`test\` merge in \`65aee7c\`):

- **\`lua/ezui/screen.lua:639\`** -- the toast \`Alt+ENTER\` handler at line 610 now checks \`services.lockscreen.is_locked()\` before firing \`t.action.on_press()\` and swallows the chord while locked, so DM-open / OTA-reboot toasts can't bypass the lockscreen's input gate.
- **\`lua/services/lockscreen.lua:262\`** -- added \`lockscreen.reset_cooldown()\` (clears \`lock_until_ms\` without touching the fail count), and call it from \`lua/boot.lua\` right before the boot-time \`lock.lock()\`. The retry-wait timer now starts fresh on every boot, matching the documented intent in \`lua/docs/manual/settings/lockscreen.md\`. The fail count survives across reboots so a power-cycle attack still pays the backoff once the user starts typing again.

\`pio run\` clean. Hardware-sensitive surface (boot-time lockscreen entry, toast handling above the screen stack); please flash and re-test the boot-lock / wrong-PIN-then-reboot / DM-toast-while-locked flows before merging.

xtr-bas:
Addressed the inline review feedback in \`5901c6e\` (preceded by a \`test\` merge in \`4f360c9\`):

- **\`lua/services/lockscreen.lua:219\`** (binary blob truncated by NVS string storage) -- the packed secret has \`\\0\\0\` bytes in the iteration count (\`ITERS = 50000\` -> \`0x50 0xC3 0x00 0x00\`), and \`ez.storage.set_pref\` uses NUL-terminated C-string semantics. Hex-encoding the blob before \`set_pref\` and decoding on read keeps the stored value all-ASCII so \`Preferences::putString\` doesn't silently truncate. No change to the wire / on-disk layout described by the blob itself -- only the NVS representation.

\`pio run\` clean. Hardware-sensitive (boot-time unlock; flash and re-test the setup -> reboot -> unlock flow with a real PIN and a real passphrase before merging). NOTE: any device that previously ran the broken code has a corrupted \`lock_secret\` and must clear it (Settings -> Lockscreen -> Off, then re-setup) before this fix takes effect.

xtr-bas:
Addressed all three new inline asks in \`d1eceda\` (preceded by a clean \`test\` merge in \`06dfb2e\`):

- **\`lua/screens/tools/lockscreen.lua:83\`** (PIN dot indicator frozen) -- \`text_input.on_change\` now calls \`self:set_state({ input = v })\` instead of mutating \`state.input\` directly, so each keystroke triggers a rebuild and the dot count updates live.
- **\`lua/screens/tools/lockscreen.lua:1\`** (touch events bypass the lock) -- extended \`lua/ezui/touch_input.lua\`'s \`M.is_locked()\` to also return \`true\` when \`services.lockscreen.is_locked()\`, mirroring the existing \`services.input_lock\` integration. Direct \`touch/*\` subscribers that already opt in (per CLAUDE.md's "Touch input and the screensaver wake gate") now get gated by the session lock too.
- **\`lua/ezui/screen.lua:714\`** (CLAUDE.md violation: bare Alt+L for a system-level global chord) -- chord changed to \`Shift+Alt+K\` (K to dodge the input-lock toggle's \`Shift+Alt+L\` / \`Shift+Alt+U\`). Updated the on-device hint text in \`lockscreen_settings.lua\` and added a new row to \`lua/docs/manual/key-bindings.md\`.

\`pio run\` clean. Hardware-sensitive (keyboard chord path + touch gate); please flash and verify: live dot updates while typing, that screens beneath the lockscreen don't fire touch handlers while locked, and that Shift+Alt+K locks from desktop / chat / any non-locked screen.

Issue #133, opened 2026-05-13
Consider embedding vector-like map data instead of rasterized tiles


bvdaakster:
## Idea

Today the map pipeline rasterizes PMTiles vector data into 64x64 indexed-color tiles (TDMAP v6: 3-bit indexed pixels, zlib-compressed, sorted index for binary search). The on-device renderer just blits these tiles and overlays labels.

Rasterizing throws away a lot of structure that we then can't get back on-device:

- Zooming requires storing the same area at multiple zoom levels — duplicated bytes for the same roads / coastlines.
- We can't restyle without regenerating the archive (the palette swap for dark/light themes works only because we kept the semantic index per pixel — a happy accident).
- Anti-aliased / sub-pixel-accurate rendering is impossible; lines look chunky at low zoom and pixellated at high zoom.
- Rotating the map (e.g. heading-up navigation mode) would be ugly with raster tiles and trivial with vectors.

## Proposal

Investigate a "vector-ish" TDMAP variant where we store, per region:

- **Polylines** for roads / railways / coastlines / waterways, quantized to e.g. 16-bit fixed-point lat/lon offsets from a tile origin
- **Polygons** for water / parks / built-up areas (same encoding)
- **Feature class** (semantic index, same 0..7 as today) per geometry so it picks up `ezui.theme.map_palette()` colors
- **Min/max zoom hints** per geometry so the renderer can cull cheaply
- Existing label table stays as-is

The on-device renderer would draw with the existing line / fill_poly primitives. We already do a fair bit of polygon rasterization in fonts / icons, so the primitive cost should be manageable.

## Open questions

- **Size vs. raster**: raster tiles compress *really* well at this resolution. Vectors win at large coverage areas (no per-zoom duplication) but might lose for dense urban tiles. Worth a back-of-envelope size comparison on a representative region before committing.
- **Render cost**: rasterizing N polylines per frame on an ESP32-S3 at 30 FPS is the main risk. May need a coarse spatial index (R-tree, or even just a uniform grid) inside each archive to keep per-frame geometry count bounded.
- **Pipeline complexity**: the converter gets harder (geometry simplification per zoom level, Douglas-Peucker or similar). This partially undoes the simplification work in the sister issue, so the two should be sequenced — simplify first, then revisit format.
- **Format compatibility**: would be a TDMAP v7, not a v6 extension. Reader gains a branch on version; writer emits one format at a time.

## Acceptance for the investigation

A short write-up (or spike branch) with:

[...]

PR #171, opened 2026-05-15
feat(maps): simplify pipeline + TDMAP v7 vector format (closes #132 #133)

### LINKED ISSUE #132: Simplify the map generation pipeline
## Problem

The current map generation pipeline under `tools/maps/` is a multi-stage tower that is hard to approach for first-time contributors:

- `pmtiles_to_tdmap.py` — converter entry point
- `config.py` — tile sources, region presets, semantic-index → grayscale lookup
- `process.py` — grayscale conversion, dithering, RLE / zlib compression
- `archive.py` — TDMAP writer / reader / inspector
- `land_mask.py` — Natural Earth land polygon download
- Plus a checkpoint/resume layer, a `viewer.html`, and a `requirements.txt`

To produce a `.tdmap` a user has to: pick a PMTiles source, install Python deps, possibly fetch a land mask, run the converter with bounds + zoom flags, then copy the output to `/sd/maps/`. Failure modes (wrong zoom range, bounds that produce no tiles, semantic index mismatch with the on-device palette) tend to surface as silent empty maps rather than clear errors.

## Goal

Cut the number of moving parts and concepts a user has to learn to make a map. Suggestions to explore:

- **One-command UX**: `tools/maps/make_map.py <region>` (or similar) that resolves source, bounds, zoom, and output path from a single config entry. Region presets become the primary interface; raw PMTiles + bounds become an advanced override.
- **Collapse internal modules**: `process.py` + `archive.py` + `land_mask.py` mostly form one pipeline. They can plausibly merge into a single `tdmap.py` module with clearly named functions, removing a few cross-file imports.
- **Drop or hide the land mask step** for the common case — bake it into the default pipeline, or skip it entirely when the PMTiles source already includes land polygons.
- **Better error surfaces**: fail loudly when bounds produce zero tiles, when zoom range is outside what the source has, or when the semantic index lookup disagrees with `ezui.theme.map_palette()` (which is the actual source of truth for colors on-device).
- **Resume / checkpoint**: still useful for large regions, but should be invisible by default — no user-facing checkpoint files left lying around for small regions that finish in under a minute.
- **Docs**: the README inside `tools/maps/` (if any) and the CLAUDE.md "Map Tools" section should shrink to a "make this region" + "advanced: roll your own" shape, not a tour of five files.

## Non-goals

- Changing the on-disk TDMAP v6 format. The reader stays as-is.
- Changing the on-device renderer or palette story.

## Acceptance

A new contributor can generate a working `.tdmap` for a named region with a single command and zero reading of internal modules. The number of files under `tools/maps/` that they need to be aware of drops to one or two.

xtr-bas:
## Summary

Closes #132 (simplify map pipeline) and #133 (vector map data).

End-to-end rebuild of the offline-map stack:
- **Source step**: drop Planetiler / Docker / mapbox-vector-tile. `make_map.py` reads OSM PBF directly via pyosmium 4.x, with automatic Geofabrik fetch for region presets. Monaco builds in ~0.6 s end-to-end.
- **Format**: TDMAP v6 (rasterized 3-bit tiles) → v7 (vector polylines + filled polygons, delta-encoded vertices, uniform-grid spatial index, zlib per record). One archive serves every zoom level at full fidelity; themes swap colours per frame without invalidating any cache.
- **On-device renderer**: rewrites `lua/ezui/widgets/map_view.lua` to draw vectors from the spatial index every frame. Two new C++ bindings (`ez.display.fill_polygon`, `ez.display.draw_polyline`) move the inner loop into native code (ear-clipping + batched line draws).
- **Pipeline simplification**: 5 Python modules (`pmtiles_to_tdmap.py`, `archive.py`, `process.py`, `land_mask.py`, `render.py`) collapse into 3 (`make_map.py`, `tdmap.py`, `regions.py`). One-command UX: `make_map.py <region>` covers the common case; `custom <pbf>` covers the advanced case.

Net **-1311 lines** across the PR.

## What changes for users

- `pip install -r tools/maps/requirements.txt` now pulls just `osmium` (no Pillow, numpy, shapely, pmtiles, mapbox-vector-tile).
- `python make_map.py netherlands` auto-downloads the Geofabrik PBF on first run and builds a `.tdmap` in seconds-to-minutes (was 20-35 minutes through Planetiler).
- Pre-v7 `.tdmap` archives no longer load — explicit "regenerate with the current writer" error rather than a silent blank map. Hard cut per discussion: there are no users yet.

## Format

TDMAP v7 layout fully documented under `CLAUDE.md › Map tools › TDMAP format (v7)`. Same 8-class semantic schema as v6 (Land / Water / Park / Building / RoadMinor / RoadMajor / Highway / Railway), same `ezui.theme.map_palette()` colour table; just stored as geometry instead of raster pixels.

## Verification

- End-to-end on-device test with a hand-built 408-byte synthetic archive: polygons fill, polylines stroke, labels render, palette respects the active theme. Screenshots in commit `dd5d4b2`.
- Round-trip writer→reader→inspect on host.
- pyosmium pipeline build of Monaco (real OSM PBF): 864 ways + 1667 areas → 2531 geometries → 110 KB archive in 0.6 s.
- `pio run` clean (60% flash).

Pushing a real Netherlands-scale archive to the device for visual verification depends on `tools/remote/ez_remote.py` handling > 16 KB transfers reliably — the existing framing protocol has a serial-FIFO drift issue under interleaved mesh traffic that's out of scope here. The synthetic-archive verification covers the on-device code paths; for production a `tdmap.py inspect` on the host is a strong sanity check.

## Test plan

- [ ] `pip install -r tools/maps/requirements.txt` succeeds in a clean venv
- [ ] `python tools/maps/make_map.py --list` shows the preset catalogue
- [ ] `python tools/maps/make_map.py monaco` completes in < 5 s and produces a v7 archive
- [ ] `python tools/maps/tdmap.py inspect monaco.tdmap` shows v7 with non-zero geometries
- [ ] On a device flashed with this branch, the Map screen opens existing-format archives with a "regenerate" message (no crash)
- [ ] Open `tools/maps/viewer.html` in a browser, drag in a `.tdmap`, polygons + polylines render and pan/zoom works


xtr-bas:
Closing per offline discussion: the v7 vector format runs into a structural perf wall on the device — country-scale archives produce a 167 MB index that can't be loaded in usable time over the T-Deck's ~357 KB/s SD interface, and per-frame viewport queries flood AsyncIO. Returning to the rasterized format on a fresh branch from test to address #132 (pipeline simplification) without the format change. #133 closed separately.

PR #172, opened 2026-05-15
feat(map): observed-coverage overlay using per-peer RSSI samples

### LINKED ISSUE #121: feat(map): range estimator overlay using observed RSSI/SNR
## Problem

Users see repeaters as advert pins on the map but have no intuition for "will my packet actually reach that one from here?" The radio stack already records RSSI and SNR for every received packet -- we just don't visualise it.

## Proposal

A toggleable overlay on the map that draws a coarse coverage ring around each repeater (or any known peer with a recent location), sized by observed link quality.

### Data

- For each known peer with a location: keep a small ring buffer of the last N (start with 16) RSSI samples taken when the peer's signal was received.
- Smooth (median, not mean -- robust to outliers).
- Map the smoothed value to a coarse "good / marginal / poor" bucket. The actual distance the link spans is unknowable without two-way ranging, so the radius represents the user's empirical confidence -- it's the distance from the current GPS fix to the peer, scaled by signal quality.

### Rendering

- The map widget's existing `overlay_fn` hook (`lua/ezui/widgets/map_view.lua:232`) draws three concentric rings per peer:
  - Inner solid ring: distance peer-to-here, when we've heard them strongly.
  - Middle dashed ring: estimated marginal radius.
  - Outer dotted ring: estimated poor-quality radius.
- Toggled on/off from the map's M-key menu.
- Color uses the accent palette so it's readable in both dark and light themes (CLAUDE.md theming rule -- use `theme.color()` not raw RGB565).

### Honest about limits

This is a heuristic -- it cannot predict where the user can reach, only describe where they have reached. Label the overlay "Observed coverage" not "Predicted coverage". Hide the rings for any peer with fewer than 4 samples in the last hour.

## Out of scope

- Path-loss modelling (terrain / Fresnel zones). Out of scope on a T-Deck -- send people to a host-side tool if they want that.
- Coverage history visualisation across days.

## Files

- `lua/services/link_quality.lua` (new) -- ring buffers per peer.
- `lua/ezui/widgets/map_view.lua` -- nothing changes; overlay is a caller concern.
- `lua/screens/tools/map.lua` -- toggle + overlay function.

xtr-bas:
## Summary

Adds an opt-in map overlay that draws rings around each known peer
(repeater, room server, contact with a saved location) sized by the
empirical distance from the user's current GPS fix to the peer, styled
by the observed RSSI bucket:

- **Solid** ring for a *good* link (>= -90 dBm), plus extrapolated
  **dashed** / **dotted** outer rings as a "where you might reach"
  envelope.
- **Dashed** ring only for a *marginal* link (-90..-110 dBm) -- we
  cannot honestly project further.
- **Dotted** ring only for a *poor* link (below -110 dBm).

The overlay is anchored at the user's current GPS fix, so without a
fix the rings simply do not paint -- the "distance peer-to-here"
heuristic would have no meaning otherwise.

A new `services/link_quality` module subscribes once to
`mesh/node_discovered` and keeps a 16-sample, in-memory ring buffer
per `pub_key_hex`. The overlay reads
`link_quality.get_quality(pub_key_hex)` and skips peers with fewer
than four samples inside the last hour, so a single stale ADVERT
cannot paint a misleading ring.

The toggle lives behind Alt+M -> "Show observed coverage" /
"Hide observed coverage" (NVS key `map_coverage`, off by default).

## Notes on design

- **In-memory only.** RSSI describes a packet, not a node, so
  persisting across reboots would just leak stale numbers. Mirrors
  the Node Store's deliberate refusal to persist `lastRssi`.
- **Labeled "observed", not "predicted".** The issue is explicit
  that this is a heuristic. The doc text and UI strings reflect
  that.
- **Dashed/dotted rings** are emulated by walking the perimeter and
  drawing every Nth pixel (`fill_rect 1x1`). The display bindings
  ship only a solid `draw_circle`; this keeps things in pure Lua
  without a new binding.
- Service init slot 15 in `boot.lua`, between `gps_track` and
  `power`. CLAUDE.md service list and `lua/docs/manual/maps.md`
  updated in the same commit.

## Test plan

- [x] `pio run` builds clean (firmware: 62.2% flash).
- [x] `luac32 -p` parses all touched Lua files.
- [ ] **Needs hardware QA**: flash to a T-Deck, enable the overlay
      from Alt+M, confirm rings render around recently-heard peers
      with locations and a current GPS fix. Confirm rings disappear
      when the toggle is off, when GPS has no fix, and when no peer
      has 4+ samples in the last hour.
- [ ] **Needs hardware QA**: verify ring style maps correctly to
      observed RSSI -- watch a single peer's RSSI move across the
      -90 / -110 dBm boundaries and confirm the inner ring switches
      between solid / dashed / dotted.

Closes #121

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -237,7 +237,15 @@ Services init in order in `lua/boot.lua`:
     from the Map screen on demand; only the reaper is wired here.
     See the GPS track recordings section below for the on-disk
     format.
-15. **power** -- 30 s battery poll that transitions between Normal /
+15. **link_quality** -- subscribes once to `mesh/node_discovered` and
+    keeps a 16-sample RSSI ring buffer per `pub_key_hex`. The Map
+    screen's "Observed coverage" overlay (issue #121, toggled via
+    Alt+M) reads this via `link_quality.get_quality(pub_key_hex)`
+    to size the rings drawn around each peer. In-memory only --
+    RSSI describes a packet, not a node, so persisting across reboots
+    would leak stale numbers (same reasoning as the Node Store's
+    refusal to persist `lastRssi`).
+16. **power** -- 30 s battery poll that transitions between Normal /
     Frugal / Survival tiers with hysteresis. Other services (`gps`,
     `ntp`, `custom_packets`) consult `power.gps_sync_allowed()` /
     `power.ntp_allowed()` / `power.allow_non_dm()` predicates rather
--- a/lua/boot.lua
+++ b/lua/boot.lua
@@ -602,6 +602,12 @@ local function boot_sequence()
         gps_track.reap_unfinalised()
     end)
 
+    -- Per-peer RSSI ring buffer (issue #121). Subscribes to
+    -- mesh/node_discovered; the map's coverage overlay reads it via
+    -- link_quality.get_quality(pub_key_hex). Passive; in-memory only.
+    local link_quality = require("services.link_quality")
+    link_quality.init()
+
     -- Power policy: poll battery every 30 s and back off radio / GPS /
     -- NTP / display in two tiers as the battery drains. See
     -- lua/services/power.lua for the truth table. Starts after the
--- a/lua/docs/manual/maps.md
+++ b/lua/docs/manual/maps.md
@@ -97,13 +97,14 @@ power toggle, so disabling GPS mid-session simply stops sampling.
 ## Observed coverage overlay
 
 The Map screen can draw coarse rings around each known peer (repeaters,
-room servers, contacts) sized by the radio signal you have actually
-heard from them. Toggle it from Alt+M -> "Show observed coverage" /
-"Hide observed coverage".
+room servers, contacts) at the empirical distance from your current
+GPS fix to the peer, styled by the radio signal you have actually
+heard from that peer. Toggle it from Alt+M -> "Show observed coverage"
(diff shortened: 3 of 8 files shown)


xtr-bas:
Addressed all three inline asks in `720fa18`:

- **`lua/screens/tools/map.lua:752`** — rings now centre on the peer (`cx, cy = ppx, ppy`) instead of the user's GPS dot. Added a comment explaining why so the next reader doesn't re-introduce the bullseye.
- **`lua/services/link_quality.lua:130` (root cause in C++)** — went with fix #2: extracted the bus-publish into `postNodeDiscoveredBus(const NodeInfo&)` in `mesh_bindings.cpp` + new `mesh_bindings.h`, and call it from `main.cpp`'s default `setNodeCallback` so `mesh/node_discovered` fires on every ADVERT unconditionally. The deprecated `ez.mesh.on_node_discovered` registration path now goes through the same helper for parity. As you noted, this also unblocks the pre-existing dead subscription in `services.contacts`.
- **`lua/screens/tools/map.lua:671`** — removed `haversine_m` and retuned the menu subtitle ("Ring at peer-to-here distance, styled by signal") and the `lua/docs/manual/maps.md` prose to match. The function had no remaining call site; ring size is the geographic user-to-peer distance, not an RSSI-derived radius.

`pio run` clean.

